PR #11, opened 2021-08-09
wip image pull secret

cppforlife:


pkg/sharing/dockerconfigjson.go
```diff
@@ -0,0 +1,46 @@
+package sharing
+
```
joe-kimmel-vmw, line 2: could supply a GoDoc along the lines of 
"implementation for dockerconfigjson secret type, see https://kubernetes.io/docs/concepts/configuration/secret/#secret-types "

cmd/controller/main.go
```diff
@@ -124,6 +127,32 @@ func registerCtrl(desc string, mgr manager.Manager,
 	return ctrl, nil
 }
 
+type reconcilerWithWatches interface {
+	reconcile.Reconciler
+	AttachWatches(controller controller.Controller) error
+}
+
+func registerCtrlMin(desc string, mgr manager.Manager,
```
joe-kimmel-vmw, line 135: question on this --  Eli and are reading and the refactor of this fn makes sense, but we can't figure out what the `Min` in the name means.    Is it important?  is there another way to say it?

pkg/sharing/secret_export_reconciler.go
```diff
@@ -47,11 +53,35 @@ func (r *SecretExportReconciler) mapSecretToExport(a handler.MapObject) []reconc
 	}
 }
 
+func (r *SecretExportReconciler) WarmUp() error {
```
joe-kimmel-vmw, line 56: not totally intuitive what purpose warmup is serving so good candidate for a GoDoc --   gist of it is that otherwise we'd be event driven and wouldn't have pre-reconciled existing resources?    (there's a question somewhere in here along the lines of,  "but isn't kubernetes level-triggered?" )

pkg/sharing/secret_exports.go
```diff
@@ -0,0 +1,98 @@
+package sharing
```
joe-kimmel-vmw, line 1: godoc again to explain "sharing". --  Sharing is Caring?  who is sharing what with whom?

pkg/sharing/secret_exports.go
```diff
@@ -0,0 +1,98 @@
...
+	exportedSecretsLock sync.RWMutex
+	exportedSecrets     map[string]exportedSecret
+}
+
+type exportedSecret struct {
+	*sgv1alpha1.SecretExport
+	*corev1.Secret
+}
+
+func (p exportedSecret) Key() string {
```
joe-kimmel-vmw, line 23: for the receivers here, `p` is a little oPaque since it's a middle letter...  feelings on `es`?  or `eSecret` if you want to sound like a dotcom era company?

pkg/sharing/secret_exports.go
```diff
@@ -0,0 +1,98 @@
...
+	if p.SecretExport.Namespace == "" {
+		panic("Internal inconsistency: missing namespace")
```
joe-kimmel-vmw, line 25: should we debate `panic` v.s. errors ?  I think this is getting called in the reconcile loop and do we really want to panic during runtime or only during initialization?  If there is a busted resource is there a way/reason for us to skip it and continue to the other(s) vs. bailing entirely?

pkg/sharing/secret_exports.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	if p.SecretExport.Name == "" {
+		panic("Internal inconsistency: missing name")
+	}
+	return p.SecretExport.Namespace + "/" + p.SecretExport.Name
+}
+
+func (p exportedSecret) Matches(matcher SecretMatcher) bool {
+	if matcher.Subject != "" {
+		// TODO we currently do not match by subject
```
joe-kimmel-vmw, line 35: should we at least log a warning here?

pkg/sharing/secret_exports.go
```diff
@@ -0,0 +1,98 @@
...
+	return false
+}
+
+func NewSecretExports(log logr.Logger) *SecretExports {
+	return &SecretExports{log: log, exportedSecrets: map[string]exportedSecret{}}
+}
+
+func (se *SecretExports) Export(export *sgv1alpha1.SecretExport, secret *corev1.Secret) {
+	if export.Namespace != secret.Namespace || export.Name != secret.Name {
+		panic("Internal inconsistency: export and secret names do not match")
```
joe-kimmel-vmw, line 62: same Q. re. panic vs error

pkg/sharing/secret_exports.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	se.exportedSecretsLock.Lock()
+	defer se.exportedSecretsLock.Unlock()
+
+	exportedSec := exportedSecret{export, secret}
+	se.exportedSecrets[exportedSec.Key()] = exportedSec
+}
+
+func (se *SecretExports) Unexport(export *sgv1alpha1.SecretExport) {
```
joe-kimmel-vmw, line 72: the `se` of our dreams has been realized!

pkg/sharing/secret_exports.go
```diff
@@ -0,0 +1,98 @@
...
+	delete(se.exportedSecrets, exportedSecret{export, nil}.Key())
+}
+
+type SecretMatcher struct {
+	Namespace  string
+	Subject    string
+	SecretType corev1.SecretType
+}
+
+func (se *SecretExports) MatchedSecretsForImport(matcher SecretMatcher) []*corev1.Secret {
```
joe-kimmel-vmw, line 85: could add a GoDoc here describing the usecase: are we importing credentials and looking for exported secrets that match?

pkg/sharing/secret_reconciler.go
```diff
@@ -0,0 +1,178 @@
+package sharing
```
joe-kimmel-vmw, line 1: in general,  ask your doc if godocs are right for each function or exported type below :)

pkg/sharing/secret_reconciler.go
```diff
@@ -0,0 +1,178 @@
...
+		return fmt.Errorf("Watching secrets: %s", err)
+	}
+
+	return controller.Watch(&source.Kind{Type: &sgv1alpha1.SecretExport{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(r.mapSecretExportToSecret),
+	})
+}
+
+func (r *SecretReconciler) mapSecretExportToSecret(a handler.MapObject) []reconcile.Request {
+	// TODO expensive call on every secret export update (no cached client used, etc)
```
joe-kimmel-vmw, line 50: obligatory - this is the TODO you already called out

pkg/sharing/secret_reconciler.go
```diff
@@ -0,0 +1,178 @@
...
+		SecretNames: r.statusSecretNames(secrets),
+	}
+
+	return r.updateSecret(secret, status, originalSecret)
+}
+
+func (r *SecretReconciler) updateSecret(secret *corev1.Secret, status SecretStatus,
+	originalSecret *corev1.Secret) (reconcile.Result, error) {
+
+	const (
```
joe-kimmel-vmw, line 136: given that we're keeping our consts hyperlocal,  is it reasonable to use the single line form rather than the paren-blocks?

cmd/controller/main.go
cppforlife: changed to registerCtrlMinimal

pkg/sharing/dockerconfigjson.go
cppforlife: added a comment.

pkg/sharing/secret_exports.go
cppforlife: done.

pkg/sharing/secret_reconciler.go
cppforlife: this is more of my own stylistic choice -- i always use var and const multiline declaration.

pkg/sharing/secret_reconciler.go
cppforlife: will be addressed in future PRs

cppforlife:
merged.

PR #17, opened 2021-08-18
More test backfilling

cppforlife:


pkg/sharing/secret_export_reconciler.go
```diff
@@ -110,33 +113,34 @@ func (r *SecretExportReconciler) Reconcile(request reconcile.Request) (reconcile
...
-	// saving the status helps trigger a cascade so that the Secrets reconciler will also respond if needed
+	// Saving the status helps trigger a cascade so that
+	// the Secrets reconciler will also respond if needed
 	defer r.updateStatus(secretExport)
 
 	return status.WithReconcileCompleted(r.reconcile(secretExport, log))
 }
 
-// reconcile looks for the Secret corresponding to the SecretExport Request that we're reconciling.
+// Reconcile looks for the Secret corresponding to the SecretExport Request that we're reconciling.
```
joe-kimmel-vmw: i think it was correct for this to be lowercase `r` reconcile.

PR #26, opened 2021-08-30
unit test leverages controller-runtime/fake/client

joe-kimmel-vmw:


pkg/sharing/secret_reconciler_test.go
```diff
@@ -0,0 +1,164 @@
...
+	placeholderSecretReloaded := &corev1.Secret{}
+	err = f8CtrlCli.Get(context.Background(), secretRequestPlaceholder.NamespacedName, placeholderSecretReloaded)
+
+	assert.Equal(t, sourceSecret.Data[".dockerconfigjson"], placeholderSecretReloaded.Data[".dockerconfigjson"])
+	assert.NotNil(t, placeholderSecretReloaded.ObjectMeta.Annotations["secretgen.carvel.dev/status"])
+	var observedStatus map[string]interface{}
+	err = json.Unmarshal([]byte(placeholderSecretReloaded.ObjectMeta.Annotations["secretgen.carvel.dev/status"]), &observedStatus)
+	assert.NoError(t, err)
+
+	// TODO: thoughts from others on the least ugly way of asserting status? these are both pretty ugly...
```
joe-kimmel-vmw: Curious if anyone has opinions here re. different options/style for these assertions:   IMO the full behavior / observability / debug-ability includes these status updates so we should assert them,  but because they're JSON encoded the assertions are a little ugly.    Is there anything that strikes you as easier / harder to read and maintain?

danielhelfand: Would just asserting once on the whole expected object make sense rather than several small assertions?

joe-kimmel-vmw: lol,    you and DK agree!  :).  appreciate the guidance!

pkg/sharing/secret_reconciler_test.go
```diff
@@ -0,0 +1,321 @@
...
+		secretExports := sharing.NewSecretExportsWarmedUp(sharing.NewSecretExports(testLogr))
+		k8sClient = fakeClient.NewFakeClient(objects...)
+		secretExportReconciler = sharing.NewSecretExportReconciler(k8sClient, secretExports, testLogr)
+		secretReconciler = sharing.NewSecretReconciler(k8sClient, secretExports, testLogr)
+		secretExports.WarmUpFunc = secretExportReconciler.WarmUp
+
+		return
+	}
+
+	t.Run("happy path", func(t *testing.T) {
```
danielhelfand: Can we have a more descriptive title than happy path? `happy path: placeholder secret status updated after ReconcileSucceeded` or something similar?

joe-kimmel-vmw: yes we can

pkg/sharing/secret_export_reconciler.go
```diff
@@ -167,7 +167,8 @@ func (r *SecretExportReconciler) reconcile(ctx context.Context, secretExport *sg
 }
 
 func (r *SecretExportReconciler) updateStatus(ctx context.Context, secretExport sg2v1alpha1.SecretExport) error {
-	r.log.Info("Updating secret export")
+	log := r.log.WithValues("secretExport Name", secretExport.Name, "status", secretExport.Status)
```
cppforlife: updateStatus should take log variable from Reconcile method (it would already have appropriate values attached)

pkg/sharing/secret_export_reconciler.go
```diff
@@ -167,7 +167,8 @@ func (r *SecretExportReconciler) reconcile(ctx context.Context, secretExport *sg
...
+	log.Info("Updating secretExport")
```
cppforlife: nit: `Updating SecretExport status`

pkg/sharing/secret_reconciler.go
```diff
@@ -161,6 +161,7 @@ func (r *SecretReconciler) reconcile(ctx context.Context, secret, originalSecret
 
 func (r *SecretReconciler) updateSecret(ctx context.Context, secret corev1.Secret, status SecretStatus,
 	originalSecret corev1.Secret) (reconcile.Result, error) {
+	log := r.log.WithValues("secret namespace", secret.Namespace, "secret name", secret.Name)
```
cppforlife: should be able to pass the log instance to updateSecret from Reconcile

pkg/sharing/secret_reconciler.go
```diff
@@ -181,9 +182,11 @@ func (r *SecretReconciler) updateSecret(ctx context.Context, secret corev1.Secre
 		return reconcile.Result{}, nil
 	}
 
+	log.Info("updating secret", "status", string(encodedStatus))
 	// TODO bother to retry to avoid having to recalculate matched secrets?
 	err = r.client.Update(ctx, &secret)
 	if err != nil {
+		log.Info("failed updating secret", "error", err.Error())
```
cppforlife: already returning error, so it will be logged by the controller runtime

pkg/sharing/secret_reconciler.go
```diff
@@ -224,7 +227,7 @@ func (e *enqueueSecretExportToSecret) Update(evt event.UpdateEvent, q workqueue.
 	typedExportOld, okOld := evt.ObjectOld.(*sg2v1alpha1.SecretExport)
 	typedExportNew, okNew := evt.ObjectNew.(*sg2v1alpha1.SecretExport)
 	if okOld && okNew && reflect.DeepEqual(typedExportOld.Status, typedExportNew.Status) {
-		e.Log.Info("Skipping SecretExport update since status did not change")
+		e.Log.Info("Skipping SecretExport update since status did not change", "SecretExport Name", typedExportOld.Name)
```
cppforlife: should do something to log namespace/name pair in hopefully similar format inside reconciler

pkg/sharing/secret_reconciler_test.go
```diff
@@ -0,0 +1,321 @@
...
+		assert.NotEqual(t, placeholderSecret1.Data[".dockerconfigjson"], placeholderSecret2.Data[".dockerconfigjson"])
+	})
+
+}
+
+func Test_SecretReconciler_updatesStatus(t *testing.T) {
+	sg2v1alpha1.AddToScheme(scheme.Scheme)
+	testLogr := zap.New(zap.UseDevMode(true))
+
+	resourcesUnderTest := func() (sourceSecret corev1.Secret, placeholderSecret corev1.Secret, secretExport sg2v1alpha1.SecretExport) {
```
cppforlife, line 142: im still thinking on how to improve definition of this. may be we can screw around with it on zoom together next week.

pkg/sharing/secret_reconciler_test.go
```diff
@@ -0,0 +1,321 @@
...
+
+		assert.Equal(t, sourceSecret.Data[".dockerconfigjson"], placeholderSecret1.Data[".dockerconfigjson"])
+		assert.Equal(t, originalPlaceholder2Data, placeholderSecret2.Data[".dockerconfigjson"])
+		assert.NotEqual(t, placeholderSecret1.Data[".dockerconfigjson"], placeholderSecret2.Data[".dockerconfigjson"])
+	})
+
+}
+
+func Test_SecretReconciler_updatesStatus(t *testing.T) {
+	sg2v1alpha1.AddToScheme(scheme.Scheme)
```
cppforlife, line 139: i think add to schema should happen in an init() for the test package. otherwise each test keeps on adding same thing.

pkg/sharing/secret_export_reconciler.go
joe-kimmel-vmw: i've put the status in as a logged kvp since we're not adding it to the context in the earlier method anymore.

pkg/sharing/secret_reconciler.go
joe-kimmel-vmw: made an attempt - i'm wondering if there's a more expedient `.NamespacedName` than calling `fmt.Sprintf` to build one but i didn't see it.

pkg/sharing/secret_reconciler_test.go
joe-kimmel-vmw: cool -  it seems like it's too much for the table-driven tests stuff but maybe it could be made to fit.     another idea i've been playing with is the easiest or "best" way to just shotgun yaml in, so that the 'fixtures' definitions can live in the examples folder - it seems like there's an `client.Create` for the k8sClient ,  and there's also "consume kapp as a library"....

pkg/sharing/secret_reconciler_test.go
joe-kimmel-vmw: leaving open as it may make sense to centralize this along with instantiation of test fixture objects.

PR #28, opened 2021-09-08
unit testing for controller reconcilers (r3)

joe-kimmel-vmw:


pkg/sharing/import_secret_test.go
```diff
@@ -0,0 +1,85 @@
...
+		destinationSecret := corev1.Secret{}
+		err := k8sClient.Get(context.Background(), namespacedNameFor(&secretImport), &destinationSecret)
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "not found")
+	})
+}
+
+func secretImportFor(sourceSecret corev1.Secret) sg2v1alpha1.SecretImport {
+	return sg2v1alpha1.SecretImport{
+
```
cppforlife: nit: remove newline

pkg/sharing/placeholder_secret_test.go
```diff
@@ -0,0 +1,270 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var testLogr logr.Logger
+
+func init() {
+	sg2v1alpha1.AddToScheme(scheme.Scheme)
+	testLogr = zap.New(zap.UseDevMode(true))
+}
```
cppforlife: no need to dup this as its already done within this package

pkg/sharing/placeholder_secret_test.go
```diff
@@ -0,0 +1,270 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-secret",
+			Namespace: "test-source",
+		},
+		Data: map[string][]byte{
+			corev1.DockerConfigJsonKey: []byte(`{"auths":{"server":{"username":"correctUser","password":"correctPassword","auth":"correctAuth"}}}`),
+		},
+		Type: "kubernetes.io/dockerconfigjson",
+	}
+
```
cppforlife: nit: rm newline

pkg/sharing/placeholder_secret_test.go
```diff
@@ -0,0 +1,270 @@
...
+func placeholderReconcilers(objects ...runtime.Object) (secretExportReconciler *sharing.SecretExportReconciler, secretReconciler *sharing.SecretReconciler, k8sClient client.Client) {
+	k8sClient = fakeClient.NewFakeClient(objects...)
+	secretExports := sharing.NewSecretExportsWarmedUp(sharing.NewSecretExports(testLogr))
+	secretExportReconciler = sharing.NewSecretExportReconciler(k8sClient, secretExports, testLogr)
+	secretExports.WarmUpFunc = secretExportReconciler.WarmUp
+	secretReconciler = sharing.NewSecretReconciler(k8sClient, secretExports, testLogr)
+	return
+}
+
+// reload asks the Kubernetes runtime client to re-populate our object
```
cppforlife, line 188: this and functions below should probably be split off into their own file.

pkg/sharing/secret_reconciler.go
```diff
@@ -224,7 +224,7 @@ func (e *enqueueSecretExportToSecret) Update(evt event.UpdateEvent, q workqueue.
 	typedExportOld, okOld := evt.ObjectOld.(*sg2v1alpha1.SecretExport)
 	typedExportNew, okNew := evt.ObjectNew.(*sg2v1alpha1.SecretExport)
 	if okOld && okNew && reflect.DeepEqual(typedExportOld.Status, typedExportNew.Status) {
-		e.Log.Info("Skipping SecretExport update since status did not change")
+		e.Log.Info("Skipping SecretExport update since status did not change", "SecretExport NamespacedName", fmt.Sprintf("%s/%s", typedExportOld.Namespace, typedExportOld.Name))
```
cppforlife: use log from the function args.

pkg/sharing/placeholder_secret_test.go
joe-kimmel-vmw: `reload` specifically is only used in this test, but i did split out a `helpers_for_test.go` for common fns.     I'm open to other file names;  I was getting a compiler error complaining about multiple namespaces in the same folder if it didn't end in `_test.go`

pkg/sharing/secret_reconciler.go
joe-kimmel-vmw: this one we can't change the signature: 
```
cannot use &enqueueSecretExportToSecret{...} (type *enqueueSecretExportToSecret) as type handler.EventHandler in argument to controller.Watch:
	*enqueueSecretExportToSecret does not implement handler.EventHandler (wrong type for Update method)
```

cppforlife: all right, then something like this:

```
e.Log.WithValues("request", types.NamespacedName{Namespace: typedExportOld.Namespace, Name: typedExportOld.Name}).Info(...)
```

PR #29, opened 2021-09-13
examples: shotgun all the examples through kapp

joe-kimmel-vmw:


hack/test-examples.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -e -x -u
+
```
danielhelfand, line 4: Nit: extra space

danielhelfand (commented):
Is the intention here to run this in CI as well? Maybe add to our ./hack/test-all.sh?

PR #34, opened 2021-10-11
Create dependabot.yml

joe-kimmel-vmw:
i'm trying to have it only make security related prs: 

> update-types—use to ignore types of updates, such as semver major, minor, or patch updates on version updates (for example: version-update:semver-patch will ignore patch updates). You can combine this with dependency-name: "*" to ignore particular update-types for all dependencies. Currently, version-update:semver-major, version-update:semver-minor, and version-update:semver-patch are the only supported options. *Security updates are unaffected by this setting.*

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#ignore

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@
 
 version: 2
 updates:
-  - package-ecosystem: "" # See documentation for possible values
+  - package-ecosystem: "gomod" # See documentation for possible values
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: "*"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "" # See documentation for possible values
```
aaronshurley: I think you might have to specify the package ecosystem. Looking at [imgpkg's config](https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/.github/dependabot.yml), "gomod" should do the trick.

aaronshurley:
@joe-kimmel-vmw out of curiosity, why not track non-security dependency bumps?

.github/dependabot.yml
joe-kimmel-vmw: good catch -- i thought i had done so,  but apparently that didn't make it in somehow,   thanks for the spot

joe-kimmel-vmw:
> @joe-kimmel-vmw out of curiosity, why not track non-security dependency bumps?

i know, right?   so the thing is,   we generally don't just upgrade to keep up with the .joneses or whomever,  and we for sure don't track the k8s version closely.    Since dependabot will open N PRs at a time (default N=5),  I was concerned that we'd get low-interest minor version bumps crowding out higher priority security patches.   I'm kind of just guessing though, I haven't actually used dependabot previously

aaronshurley:
> I was concerned that we'd get low-interest minor version bumps crowding out higher priority security patches.

Gotcha, that makes sense. I'm supportive of this approach. We can always open it up further in the future if we want.

Issue #38, opened 2021-11-05
support excluding namespace from being matched by wildcards


cppforlife:
**Describe the problem/challenge you have**

currently users of SecretExport may specify `toNamespace: "*"`. there might be namespaces in the cluster that cluster administrators would want to section off from receiving any secrets unless SecretExport explicitly specifies namespace selection.

**Describe the solution you'd like**
one idea in how to mark such namespaces is via annotation (on a namespace) -- `secretgen.carvel.dev/excluded-from-wildcard-matching: ""`

danielhelfand:
Making some notes here:
* We need to update the [RBAC](https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/10089556361f662133c8e8bf0be9596f8791ef0a/config/rbac.yml#L14) to allow getting namespaces
* We should also add some [docs](https://github.com/vmware-tanzu/carvel-secretgen-controller/tree/develop/docs) that mention this annotation / how to use it

aaronshurley:
This has been delivered as a part of [v0.7.0](https://github.com/vmware-tanzu/carvel-secretgen-controller/releases/tag/v0.7.0). But, we recommended using [v0.7.1](https://github.com/vmware-tanzu/carvel-secretgen-controller/releases/tag/v0.7.1).

jorgemoralespou:
Seems as this feature got implemented and merged already. I'm not still convinced by this solution, as it means that the author of a package (let's say my training-package) would need to add this annotation to my namespace definition just to prevent from secretgen controller to inject me a secret even if I, as package author don't know whether my package is going to be deployed in a cluster where secretgen controller does exist. This means that this solution introduces a dependency on package authors to add an annotation just in case.
I would think that operators/cluster admins that are installing features on the cluster and that create a secretgen are the ones that know which capabilities exist and whether they should prevent secretgen from leaking secrets into namespaces that it shouldn't. For this reason, I think that this solution solves the problem for one specific use-case, but definitely there will be reports to implement this feature in a more sustainable manner (for package authors) in the future.

The responsibility issue can probably be solved by documenting how to create an overlay in the package to have this annotation added so that a COTS package can exclude the wildcard secretgen.

PR #41, opened 2021-11-10
support excluding namespace from being matched by wildcards

joe-kimmel-vmw:
…for the annotation to appear / disappear


pkg/sharing/secret_reconciler.go
```diff
@@ -116,6 +116,19 @@ const (
...
+			Name: secret.Namespace,
+		}
+		namespace := corev1.Namespace{}
+		err := r.client.Get(context.Background(), query, &namespace)
+		if err != nil {
+			r.log.Error(err, "predictWantToReconcile couldn't find:", "namespace", secret.Namespace)
+			return false
+		}
+		_, excluded := namespace.Annotations["secretgen.carvel.dev/excluded-from-wildcard-matching"]
+		found = !excluded
```
cppforlife: predictWantToReconcile is not a "chokepoint" for whether to reconcile something or not -- Reconcile method in theory could be called at the start of the controller on all secrets.

the main chokepoint is https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/pkg/sharing/secret_exports.go#L96 where we determine which secrets should be imported (this also affects SecretImport controller). if destination namespace is marked with above annotation, _and_ if SecretExport is matching it because of "*", then we should not include this secret. SecretExports with explicit namespace name are still allowed.

pkg/sharing/secret_reconciler.go
```diff
@@ -116,6 +116,19 @@ const (
 
 func (r *SecretReconciler) predictWantToReconcile(secret corev1.Secret) bool {
 	_, found := secret.Annotations[imagePullSecretAnnKey]
+	if found {
```
joe-kimmel-vmw: sorry, just noticed i didn't take this stuff back out again;    will remove if we're keepign the above changes, otherwise probably just start a fresh branch anyhow.

pkg/sharing/secret_exports_test.go
```diff
@@ -47,10 +47,13 @@ func TestSecretExports(t *testing.T) {
...
-			ToNamespace: "dst-ns",
-			SecretType:  corev1.SecretType("Opaque"),
-		}))
+		fakeNsCheck := func(string) bool { return false }
+		require.Equal(t, []*corev1.Secret(nil),
+			se.MatchedSecretsForImport(sharing.SecretMatcher{
+				ToNamespace: "dst-ns",
+				SecretType:  corev1.SecretType("Opaque"),
+			},
+				fakeNsCheck))
```
cppforlife: nit: formatting

pkg/sharing/secret_exports_test.go
```diff
@@ -47,10 +47,13 @@ func TestSecretExports(t *testing.T) {
 		}
 		se.Export(export2, secret2)
 
-		require.Equal(t, []*corev1.Secret(nil), se.MatchedSecretsForImport(sharing.SecretMatcher{
-			ToNamespace: "dst-ns",
-			SecretType:  corev1.SecretType("Opaque"),
-		}))
+		fakeNsCheck := func(string) bool { return false }
```
cppforlife: nit: i typically recommend against using name fake* in test. it just becomes noisy and doesnt add any value. either nsCheck or negativeExclusionCheck

pkg/sharing/secret_exports.go
```diff
@@ -93,14 +96,14 @@ type SecretMatcher struct {
 //       - secret with specific namespace
 //       - secret with wildcard namespace match
 //     (in all cases fallback to secret namespace/name sort)
-func (se *SecretExports) MatchedSecretsForImport(matcher SecretMatcher) []*corev1.Secret {
+func (se *SecretExports) MatchedSecretsForImport(matcher SecretMatcher, nsIsExcluded NamespaceExclusionCheck) []*corev1.Secret {
```
cppforlife: should we include this on SecretMatcher struct? SecretMatcher contains ToNamespace, so may be `IsNamespaceExcludedFromWildcardMatch func(string) bool` field.

pkg/sharing/secret_exports.go
```diff
@@ -172,17 +175,20 @@ func (es exportedSecret) Matches(matcher SecretMatcher, log logr.Logger) bool {
...
 
-func (es exportedSecret) matchesNamespace(nsToMatch string) bool {
+func (es exportedSecret) matchesNamespace(nsToMatch string, nsIsExcluded NamespaceExclusionCheck) bool {
 	for _, ns := range es.export.StaticToNamespaces() {
-		if ns == sg2v1alpha1.AllNamespaces || ns == nsToMatch {
+		if ns == nsToMatch {
 			return true
 		}
+		if ns == sg2v1alpha1.AllNamespaces {
+			return !nsIsExcluded(nsToMatch)
```
cppforlife: this shouldnt be an exact return. since you may have `[*, ns1]`, so even though `*` is not allowed, ns1 may still match.

pkg/sharing/reconciler_helpers.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// makeNamespaceExclusionCheck returns a function that uses reconciler-level tools (k8s client, logger, context) to
+// check the presence of a namespace annotation that we mostly only care about in the inner workings of SecretExport.
+func makeNamespaceExclusionCheck(ctx context.Context, cli client.Client, log logr.Logger) NamespaceExclusionCheck {
```
cppforlife: term `cli` is a pretty confusing thing to read in a server side world.

pkg/sharing/secret_exports.go
cppforlife: im also thinking of may be we should change "behaviour/naming" -- DoesNamespaceAcceptSecrets() bool

pkg/sharing/reconciler_helpers.go
joe-kimmel-vmw: `kubernetesClient` ?

pkg/sharing/secret_exports.go
joe-kimmel-vmw: re. putting it in the struct vs. passing as an arg -  i went back and forth and obviously settled here.  If the `ToNamespace` is already known when we make the struct there's an argument of like, "why didn't you invoke the function on the name, and then not bother making the struct at all if the check was going to fail?"


i don't think `DoesNamespaceAcceptSecrets` works,  because per your spec it _does_ accept secrets if it's explicitly named.   so it's back to `DoesNamespaceAcceptWildcardSecretExports` or something similarly out of some Java architect's BufferedReaderWriter fantasy. 

... i haven't come up with a decent short name.  not trying to defend whatever i used yesterday,  but not totally sold on the longer versions of `NsSezYesToStarShares`

pkg/sharing/secret_exports.go
joe-kimmel-vmw: i'm confused - -this function takes a string, not a []string ?

joe-kimmel-vmw:
#38

pkg/sharing/secret_import_reconciler.go
```diff
@@ -92,6 +94,43 @@ func (r *SecretImportReconciler) AttachWatches(controller controller.Controller)
 			return result
 		},
 	})
+	if err != nil {
+		return err
+	}
+
+	// Watch namespaces partly so that we cache them because we migh be doing a lot of lookups
+	return controller.Watch(&source.Kind{Type: &sg2v1alpha1.SecretExport{}}, &enqueueNamespaceToSecret{
```
cppforlife: this is watching SecretExport

pkg/sharing/secret_import_reconciler.go
```diff
@@ -237,3 +277,41 @@ func (r *SecretImportReconciler) updateStatus(
...
+// Delete doesn't do anything
+func (e *enqueueNamespaceToSecret) Delete(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
+
+}
+
+// Generic doesn't do anything
+func (e *enqueueNamespaceToSecret) Generic(evt event.GenericEvent, q workqueue.RateLimitingInterface) {
+
+}
+func (e *enqueueNamespaceToSecret) mapAndEnqueue(q workqueue.RateLimitingInterface, object client.Object) {
```
cppforlife, line 307: nit: newline above

pkg/sharing/secret_import_reconciler.go
```diff
@@ -237,3 +277,41 @@ func (r *SecretImportReconciler) updateStatus(
...
+	if okOld && okNew && (nsHasExclusionAnnotation(*typedNsOld) == nsHasExclusionAnnotation(*typedNsNew)) {
+		return // Skip when exclusion annotation did not change
+	}
+
+	e.mapAndEnqueue(q, evt.ObjectNew)
+}
+
+// Delete doesn't do anything
+func (e *enqueueNamespaceToSecret) Delete(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
+
```
cppforlife: nit: can turn into `{}`

pkg/sharing/secret_reconciler.go
```diff
@@ -51,11 +51,49 @@ func (r *SecretReconciler) AttachWatches(controller controller.Controller) error
...
 		ToRequests:    r.mapSecretExportToSecret,
 		Log:           r.log,
 	})
+	if err != nil {
+		return fmt.Errorf("Watching secretExports: %s", err)
+	}
+
+	// Watch namespaces partly so that we cache them because we migh be doing a lot of lookups
+	// note that for now we are using the same enqueueNamespaceToSecret as the secretImportReconciler
+	return controller.Watch(&source.Kind{Type: &sg2v1alpha1.SecretExport{}}, &enqueueNamespaceToSecret{
```
cppforlife: not watching Namespace

pkg/sharing/secret_reconciler.go
```diff
@@ -51,11 +51,49 @@ func (r *SecretReconciler) AttachWatches(controller controller.Controller) error
...
+		ToRequests: r.mapNamespaceToSecret,
+		Log:        r.log,
+	})
+}
+
+func (r *SecretReconciler) mapNamespaceToSecret(ns client.Object) []reconcile.Request {
+	var secretList corev1.SecretList
+	// TODO: scope the secret list to the namespace, which i think is the client.Object that got passed in...
+	err := r.client.List(context.Background(), &secretList)
```
cppforlife: is there no way to tell List to only give you namespace X objects?

pkg/sharing/placeholder_secret_test.go
```diff
@@ -52,6 +52,47 @@ func Test_SecretReconciler_respectsNamespaces(t *testing.T) {
...
+		secretExportReconciler, secretReconciler, k8sClient = placeholderReconcilers(&sourceSecret, &placeholderSecret1, &placeholderSecret2, &secretExport, &excludedNs, &includedNs)
+		reconcileObject(t, secretExportReconciler, &secretExport)
+		reconcileObject(t, secretReconciler, &placeholderSecret1)
+		reconcileObject(t, secretReconciler, &placeholderSecret2)
+
+		reload(t, &placeholderSecret1, k8sClient)
+		reload(t, &placeholderSecret2, k8sClient)
+		assert.Equal(t, sourceSecret.Data[".dockerconfigjson"], placeholderSecret1.Data[".dockerconfigjson"])
+		assert.Equal(t, sourceSecret.Data[".dockerconfigjson"], placeholderSecret2.Data[".dockerconfigjson"])
+
```
cppforlife: nit: remove newline

pkg/sharing/secret_import_reconciler.go
joe-kimmel-vmw: :(

docs/secret-export.md
```diff
@@ -100,6 +100,12 @@ Their main purpose is to provide an alternative to `SecretImport` to import comb
 
 ***Warning*** Since SecretExport CR allows you to export registry credentials to other namespaces, they will become visible to users of such namespaces. We strongly recommend to ensure that registry credentials you are exporting only allow read-only access to the registry with minimal necessary scope.
 
+***Exclude From Wildcard Matching***
+You can mark a namespace as exempt from wildcard matching (i.e. `*` exports) so
+that placeholder secrets in that ns will only be populated by secretExports that
```
danielhelfand:
```suggestion
that placeholder secrets in that namespace will only be populated by SecretExports that
```

pkg/sharing/namespace_wildcard_exclusions.go
```diff
@@ -0,0 +1,37 @@
...
+	kubernetesClient client.Client,
+	log logr.Logger) NamespaceWildcardExclusionCheck {
+	return func(nsName string) bool {
+		query := types.NamespacedName{
+			Name: nsName,
+		}
+		namespace := corev1.Namespace{}
+		err := kubernetesClient.Get(ctx, query, &namespace)
+		if err != nil {
+			log.Error(err, "Called to check annotation on a namespace but couldn't find:", "namespace", nsName)
```
danielhelfand, line 27: Is there a `log.Errorf` func equivalent for this?

pkg/sharing/secret_import_reconciler.go
```diff
@@ -92,6 +94,42 @@ func (r *SecretImportReconciler) AttachWatches(controller controller.Controller)
...
+	err := r.client.List(context.Background(), &secretList, client.InNamespace(ns.GetName()))
+	if err != nil {
+		// TODO what should we really do here?
+		r.log.Error(err, "Failed fetching list of all secrets")
+		return nil
+	}
+
+	var result []reconcile.Request
+	for _, secret := range secretList.Items {
+		if secret.Namespace == ns.GetName() {
```
cppforlife: if we List is doing ns filtering, then we shouldnt have to do it here.

pkg/sharing/secret_import_reconciler.go
```diff
@@ -13,10 +13,12 @@ import (
 	"github.com/vmware-tanzu/carvel-secretgen-controller/pkg/reconciler"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	// metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/workqueue"
+
```
danielhelfand, line 17: Can you reformat the imports here? There should be no spaces. I also generally run `goimports -w file` to make sure everything is organized appropriately.

pkg/sharing/secret_import_reconciler.go
```diff
@@ -92,6 +94,42 @@ func (r *SecretImportReconciler) AttachWatches(controller controller.Controller)
 			return result
 		},
 	})
+	if err != nil {
+		return err
+	}
+
+	// Watch namespaces partly so that we cache them because we migh be doing a lot of lookups
```
danielhelfand:
```suggestion
	// Watch namespaces partly so that we cache them because we might be doing a lot of lookups
```

pkg/sharing/secret_import_reconciler.go
```diff
@@ -92,6 +94,42 @@ func (r *SecretImportReconciler) AttachWatches(controller controller.Controller)
...
+		ToRequests: r.mapNamespaceToSecret,
+		Log:        r.log,
+	})
+}
+
+func (r *SecretImportReconciler) mapNamespaceToSecret(ns client.Object) []reconcile.Request {
+	var secretList corev1.SecretList
+	err := r.client.List(context.Background(), &secretList, client.InNamespace(ns.GetName()))
+	if err != nil {
+		// TODO what should we really do here?
```
danielhelfand, line 112: Can we remove this comment or provide more details?

pkg/sharing/secret_reconciler.go
```diff
@@ -51,11 +51,49 @@ func (r *SecretReconciler) AttachWatches(controller controller.Controller) error
...
+	// note that for now we are using the same enqueueNamespaceToSecret as the secretImportReconciler
+	return controller.Watch(&source.Kind{Type: &corev1.Namespace{}}, &enqueueNamespaceToSecret{
+		ToRequests: r.mapNamespaceToSecret,
+		Log:        r.log,
+	})
+}
+
+func (r *SecretReconciler) mapNamespaceToSecret(ns client.Object) []reconcile.Request {
+	var secretList corev1.SecretList
+	// TODO: scope the secret list to the namespace, which i think is the client.Object that got passed in...
```
cppforlife: stale comment

pkg/sharing/secret_import_reconciler.go
```diff
@@ -237,3 +276,38 @@ func (r *SecretImportReconciler) updateStatus(
...
+	}
+
+	e.mapAndEnqueue(q, evt.ObjectNew)
+}
+
+// Delete doesn't do anything
+func (e *enqueueNamespaceToSecret) Delete(evt event.DeleteEvent, q workqueue.RateLimitingInterface) {}
+
+// Generic doesn't do anything
+func (e *enqueueNamespaceToSecret) Generic(evt event.GenericEvent, q workqueue.RateLimitingInterface) {
```
cppforlife, line 304: nit: `{}` formatting

pkg/sharing/secret_import_reconciler.go
```diff
@@ -92,6 +94,42 @@ func (r *SecretImportReconciler) AttachWatches(controller controller.Controller)
...
+				NamespacedName: types.NamespacedName{
+					Name:      secret.Name,
+					Namespace: secret.Namespace,
+				},
+			})
+		}
+	}
+
+	r.log.Info("Planning to reconcile matched secrets",
+		"all", len(secretList.Items), "matched", len(result))
```
cppforlife: all and matched would always be the same

pkg/sharing/secret_reconciler.go
```diff
@@ -51,11 +51,49 @@ func (r *SecretReconciler) AttachWatches(controller controller.Controller) error
...
+	err := r.client.List(context.Background(), &secretList, client.InNamespace(ns.GetName()))
+	if err != nil {
+		// TODO what should we really do here?
```
danielhelfand, line 75: Can we remove this comment or provide more details?

pkg/sharing/placeholder_secret_test.go
```diff
@@ -52,6 +52,46 @@ func Test_SecretReconciler_respectsNamespaces(t *testing.T) {
...
+				Name: placeholderSecret1.Namespace,
+			},
+		}
+		secretExport := secretExportFor(sourceSecret, "*")
+		secretExportReconciler, secretReconciler, k8sClient := placeholderReconcilers(&sourceSecret, &placeholderSecret1, &placeholderSecret2, &secretExport, &excludedNs, &includedNs)
+		reconcileObject(t, secretExportReconciler, &secretExport)
+		reconcileObject(t, secretReconciler, &placeholderSecret1)
+		reconcileObject(t, secretReconciler, &placeholderSecret2)
+		// placeholder secret2 should have its original contents for auths and a helpful status message
+		originalPlaceholder2Data := append([]byte{}, placeholderSecret2.Data[".dockerconfigjson"]...)
```
cppforlife, line 76: i dont get why we are capturing `originalPlaceholder2Data` _after_ doing reconciles. wouldnt moving it above all the reconciler stuff make it more clear that it's not being modified by all the reconcile*

pkg/sharing/secret_import_reconciler.go
joe-kimmel-vmw: hmm. - for the record part of what happens is that some combination of things running in my editor disagree with your formatting preferences.    there's hooks for customizing gofmt, right?

joe-kimmel-vmw: like i don't how i got away with line 303 above this,   right now every time i save it just puts the `}` back on the line below

pkg/sharing/secret_import_reconciler.go
joe-kimmel-vmw: i'm not sure what you mean "no spaces" -    blank lines ?     there is definitely a process that auto-runs and reorganizes my imports ,  but i don't have `goimports` in my shell path.    but like every time i save the imports are reformatted.  i feel like there's a thread in this PR from both you and Dmitriy that's like,  somehow "i'm just using luanvim" is coming up to mean that my editor is enforcing different opinions than what you guys have?  How did this not come up previously

pkg/sharing/secret_import_reconciler.go
joe-kimmel-vmw: i can remove it but i think the detail is : "I don't know what to do here."   I copied over the comment from existing code setting up a similar watch -- someone else was unsure what to do here also. see:
https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/39cc31078f2ea13daa7aaa662d9c91308de85864/pkg/sharing/secret_import_reconciler.go#L74

and
https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/8fc61ebfb2fed4a8d7bfac2c193cafb3149dafd7/pkg/sharing/secret_reconciler.go#L67

I'm not saying it's the _most_ helpful comment,  but i think leaving it in shows that the original author didn't have informed intentions --  like it's not, "oh it must be this way for a reason". it's just "oh it's this way because we had to _something_ on the error, and `panic` seemed like the wrong answer."

pkg/sharing/placeholder_secret_test.go
joe-kimmel-vmw: oh, sure, i could move it -- might make it easier to skim but at the cost of obscuring what's actually happening:  even though you pass a pointer to the reconciler calls, they don't actually modify the input arg - eventually somebody makes a copy.

  that's why we have to reload at all - if the reconcilers were modifying the structs that we have references to we wouldn't need to reload!

joe-kimmel-vmw: see e.g. line 83 below this i tried to call out:  after we add the annotation we actually have to remake the fake-client because otherwise it'll keep returning (a copy of) the same struct that i passed in originally.

pkg/sharing/namespace_wildcard_exclusions.go
cppforlife: i think log package is meant to capture KVs instead of usual sprintf formatting

pkg/sharing/placeholder_secret_test.go
joe-kimmel-vmw: I added more comments trying to explain...

pkg/sharing/namespace_wildcard_exclusions.go
joe-kimmel-vmw: i'm not sure what you're getting at:   you don't like the "structured" part of the logging?

joe-kimmel-vmw: I think the option is "call fmt.sprintf (or similar) for yourself if you really want a formatted string"

PR #45, opened 2021-11-23
adds dev deploy script

joe-kimmel-vmw:
I tested the timing of `rapid_deploy` vs. traditional `deploy` by adding a trivial additional logging line in pkg/sharing/secret_reconciler.go which is probably a best case for the rapid deploy since most of the go compiler object-file cache was still valid. I also only did N=1 test.    nonetheless,  it's roughly an order of magnitude speedup: 

`./hack/deploy.sh  15.73s user 20.45s system 22% cpu 2:41.21 total` (~160 seconds)

(another trivial log line change to force at least a trivial recompilation and creation of a new container layer)
`./hack/rapid-deploy.sh  4.71s user 3.77s system 77% cpu 10.913 total` (~11 seconds)

"rapid" is faster partly b/c it skips the hygiene steps of `gofmt`, and  `go mod ...`, partly because it compiles only once, and partly because it compiles against a persistent file system allowing it to leverage the object file cache.    

I'm thinking a good flow would be to develop a feature and deploy it for the first time with `deploy` and then use `rapid-deploy` for iterating on small fixes.


hack/rapid-deploy.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -ex
+
+CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=develop' -buildid=" -trimpath -o linux_controller ./cmd/controller/...
```
joe-kimmel-vmw: i called the output `linux_controller` but feel free to tell me that it's not necessary and we should just continue to use the `controller` that was already in the .gitignore.

danielhelfand: Think it's helpful for users locally to see the difference. I think keeping `linux_controller` makes sense.

hack/rapid-deploy.sh
```diff
@@ -0,0 +1,14 @@
...
+
+echo "# -- this file is autogenerated by rapid-deploy.sh from main Dockerfile and is not version controlled" > Dockerfile.dev
+
+run_image_start=`cat Dockerfile | grep -n "\- run \-" | cut -d':' -f1`
+tail -n +$run_image_start Dockerfile | \
+  sed 's/COPY.*secretgen-controller/COPY linux_controller secretgen-controller/' >> Dockerfile.dev
+
+ytt -f config/ -f config-rapid-deploy/ | kbld -f- | kapp deploy -a sg -f- -c -y
```
danielhelfand: Can we maybe call this `config-dev-deploy` instead?

danielhelfand (commented):
Generally lgtm. Some nits added and some minor questions:

1. Have you run e2e tests against using this deployment workflow and did everything pass?
2. Can we add some dev docs somewhere on when to use this workflow/design of it all? Probably in [dev.md](https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/47d2f3270dbca5977d07608a99772f1f582f8877/docs/dev.md).

hack/rapid-deploy.sh
cppforlife: may be `controller-linux-amd64` a bit better name?

hack/rapid-deploy.sh
joe-kimmel-vmw: cool --  it started life as "dev-deploy" and then i thought,  "wait aren't _all_ the deploys that i do 'dev'?" so i went with "rapid" ... happy to change it back, seems DH & DK agree on the name.

joe-kimmel-vmw:
> 1. Have you run e2e tests against using this deployment workflow and did everything pass?

 yes -- actually not the whole suite but I assume if a couple of tests pass then all the other ones would too

> 2. Can we add some dev docs [...]

yes!  :)  added.

Issue #39, opened 2021-11-08
Enable trivy scan as part of nightly build


aaronshurley:
As described in vmware-tanzu/carvel-kapp-controller#368 and vmware-tanzu/carvel-kapp#315, it might be helpful for us to consider adding a security scan as part of a nightly build process. This would help us monitor and triage security alerts better for secretgen-controller. In particular, conducting a scan of the secretgen-controller image could be helpful.

**Note**: if setting up trivy is trivial, consider setting up for other carvel repos.

PR #46, opened 2021-12-14
add trivy scanning to sg-controller

joe-kimmel-vmw:
#39

### DIFF
--- a/.github/workflows/trivy-scan.yml
+++ b/.github/workflows/trivy-scan.yml
@@ -1,0 +1,126 @@
+name: Trivy CVE Dependency Scanner
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  scan-latest-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get Latest Release Docker Image Sha
+        id: latest-sha
+        run: |
+          # Get the latest released docker image sha
+          curl -sL https://api.github.com/repos/vmware-tanzu/carvel-secretgen-controller/releases/latest | jq -r '.assets[].browser_download_url' | wget -i -
+
+          echo ::set-output name=image::$(yq eval '.spec.template.spec.containers[0].image' release.yml -N -j | jq 'select(. != null)' -r)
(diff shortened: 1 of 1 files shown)


.github/workflows/trivy-scan.yml
```diff
@@ -0,0 +1,126 @@
...
+        uses: actions/setup-go@v1
+        with:
+          go-version: "1.17"
+      - name: Build the secretgen-controller artifacts
+        run: |
+          ./hack/install-deps.sh
+          ./hack/build.sh
+
+          # docker image
+          docker build -t docker.io/carvel/secretgen-controller:${{ github.sha }} .
```
joe-kimmel-vmw, line 74: I was a little surprised we're (still?) using docker.io/carvel --  was there a conversation in KC about whether to use docker.ui vs ghcr.io ?

danielhelfand: I don't think we need either `docker.io` or `ghcr.io` since this is just a local build anyway. Not sure if trivy cares about this or not.

neil-hickey: Yep, as daniel says. It's not pushing to an external registry, just locally. I don't know if this can removed, I didn't try... the trivy example had it this way, so I used that. @joe-kimmel-vmw

.github/workflows/trivy-scan.yml
```diff
@@ -0,0 +1,126 @@
...
+        with:
+          channel-id: C02MTA7F2G1
+          slack-message: "${{ env.SUMMARY }}"
+      - name: Send Failure notification
+        if: failure()
+        uses: slackapi/slack-github-action@v1.15.0
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+        with:
+          channel-id: G01FTP43JMQ
```
neil-hickey, line 53: I think the channel is not being found for some reason. I didn't investigate, I just noticed the other scanner has been failing https://github.com/vmware-tanzu/carvel-kapp-controller/runs/4554781469?check_suite_focus=true

Might be worth double checking the channel id. The bot should have permissions to post in all channels/

joe-kimmel-vmw: `G01FTP43JMQ` looks the same, no?

joe-kimmel-vmw: maybe because it's private?

PR #48, opened 2022-01-19
add an annotation to the source secret and a failing test that assert…

joe-kimmel-vmw:
…s it gets copied to the placeholder secret.

This test fails, with output:

```
placeholder Annotations:  map[secretgen.carvel.dev/image-pull-secret: secretgen.carvel.dev/status:{"conditions":[{"type":"ReconcileSucceeded","status":"True"}],"secretNames":["test-source/test-secret"]}]
source Annotations:  map[test-source-annotation:value]
    placeholder_secret_test.go:137:
        	Error Trace:	placeholder_secret_test.go:137
        	Error:      	Should be true
        	Test:       	Test_SecretReconciler_updatesStatus/one_secret_exports_successfully_to_placeholder
        	Messages:   	failed to find source key test-source-annotation in placeholder annotations
    placeholder_secret_test.go:138:
        	Error Trace:	placeholder_secret_test.go:138
        	Error:      	Not equal:
        	            	expected: "value"
        	            	actual  : ""

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-value
        	            	+
        	Test:       	Test_SecretReconciler_updatesStatus/one_secret_exports_successfully_to_placeholder
        	```

### DIFF
--- a/pkg/sharing/placeholder_secret_test.go
+++ b/pkg/sharing/placeholder_secret_test.go
@@ -6,6 +6,7 @@ package sharing_test
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/sharing/placeholder_secret_test.go [hunk 1]
```diff
@@ -128,6 +129,15 @@ func Test_SecretReconciler_updatesStatus(t *testing.T) {
 
 		reload(t, &placeholderSecret, k8sClient)
 		assert.Equal(t, sourceSecret.Data[".dockerconfigjson"], placeholderSecret.Data[".dockerconfigjson"])
+		assert.Equal(t, sourceSecret.Data[".dockerconfigjson"], placeholderSecret.Data[".dockerconfigjson"])
```
danielhelfand, line 132: Any reason for needing this line? I see the same thing on line 131.

pkg/sharing/placeholder_secret_test.go
```diff
@@ -128,6 +129,15 @@ func Test_SecretReconciler_updatesStatus(t *testing.T) {
...
+		fmt.Println("placeholder Annotations: ", placeholderSecret.ObjectMeta.Annotations)
+		fmt.Println("source Annotations: ", sourceSecret.ObjectMeta.Annotations)
+		for k, v := range sourceSecret.ObjectMeta.Annotations {
+			placeholderV, check := placeholderSecret.ObjectMeta.Annotations[k]
+			assert.True(t, check, fmt.Sprintf("failed to find source key %s in placeholder annotations", k))
+			assert.Equal(t, v, placeholderV)
```
danielhelfand, line 138: Why does `v` equal `placeholderV` here? Shouldn't the secrets have different annotations? Sorry if I am misinterpreting.

joe-kimmel-vmw: right - sorry.  i was intentionally writing a failing test. - I meant for this to be more like, 'communication through source code' than 'a change proposal'.   I put it as a "draft PR" but should've been more explicit that I don't intend to merge. 

if we wanted to merge something like this i would definitely have to make changes. :)

danielhelfand: Oh right. I should read the pr description :).

PR #52, opened 2022-02-08
SecretImports no longer copy source secret labels and annotations

### LINKED ISSUE #47: instruct secret gen controller to leave out certain annotations and labels while copying the secrets
Shreyas Sreenivas  21 hours ago
Hi,
Needed some inputs regarding the Secret-Gen-Controller SecretImport implementation and possible changes.
Secret Import duplicates the secret into a new namespace. This duplicated secret duplicates all the Annotations and Labels from the source secret (DeepCopy of the secret). This is causing an issue in the TMC managedSecret controller, this duplicate secret is getting cleaned up as the controller use annotations to map it against the server intent and the duplicate secret has the same unique identifier (TMC related ID ) in the secret annotations.

Your thoughts and suggestion on solving this problem ? (edited) 

Request for change:
Can secret gen controller on duplication of the secret add an extra annotation which depicts that the secret is created via secret-import. In the present implementation, we see that the duplicate secret does have a owner ref to secret import CR, but annotation should be simpler. By this change we can avoid the secret getting deleted from the controller

OR, can the secret gen controller avoid copying the annotations of the secrets ?

In slack thread it looked like there were two votes for the second proposal ("avoid copying certain annotations")

benmoss:
Fixes #47 

Open questions:
- Does this need to be configurable? Do any users want/rely on annotations and labels being copied? I don't see why they would.

### DIFF
--- a/pkg/reconciler/secret.go
+++ b/pkg/reconciler/secret.go
@@ -88,32 +88,8 @@ func (p *Secret) ApplyTemplate(template sgv1alpha1.SecretTemplate) error {
 	return nil
 }
 
-// ApplySecret fills in annotations, labels, type, and data on top of the Secret.
+// ApplySecret fills in type and data on top of the Secret.
 func (p *Secret) ApplySecret(otherSecret corev1.Secret) {
-	if len(otherSecret.Annotations) > 0 {
-		if p.secret.Annotations == nil {
-			p.secret.Annotations = map[string]string{}
-		}
-		for k, v := range otherSecret.Annotations {
-			// Do not override existing annotations
-			if _, found := p.secret.Annotations[k]; !found {
-				p.secret.Annotations[k] = v
-			}
-		}
-	}
-
-	if len(otherSecret.Labels) > 0 {
-		if p.secret.Labels == nil {
-			p.secret.Labels = map[string]string{}
-		}
-		for k, v := range otherSecret.Labels {
-			// Do not override existing labels
-			if _, found := p.secret.Labels[k]; !found {
-				p.secret.Labels[k] = v
-			}
-		}
-	}
-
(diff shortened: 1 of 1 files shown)


pkg/reconciler/secret.go
```diff
@@ -90,30 +90,6 @@ func (p *Secret) ApplyTemplate(template sgv1alpha1.SecretTemplate) error {
 
 // ApplySecret fills in annotations, labels, type, and data on top of the Secret.
```
joe-kimmel-vmw: :( comment is now stale

joe-kimmel-vmw:
LGTM but we should leave open to see if any of the folks in https://github.com/vmware-tanzu/carvel-secretgen-controller/issues/47 have any comments,  and also in case @cppforlife wants to veto as it does technically break backwards compatibility.

joe-kimmel-vmw:
LGTM but we should leave open to see if any of the folks in https://github.com/vmware-tanzu/carvel-secretgen-controller/issues/47 have any comments,  and also in case @cppforlife wants to veto as it does technically break backwards compatibility.

jorgemoralespou:
> Does this need to be configurable? Do any users want/rely on annotations and labels being copied? I don't see why they would.

I would think that it might be needed in some cases. @grahamdumpleton can talk to this

GrahamDumpleton:
I already commented in https://github.com/vmware-tanzu/carvel-secretgen-controller/issues/47#issuecomment-1017859278

The idea of copying existing labels/annotations was always wrong to me. In my own secret copier implementations though I do allow a copying rule to explicitly provide new labels (no annotations) that should be added to the copy as see that as useful as a new user of secret may need labels for identification purposes.

PR #65, opened 2022-05-13
Generate Package artifacts for SGC during release process

neil-hickey:
This PR creates a Package of Secretgen controller which is consumable by third parties.

It follows best practices for ytt, and can serve as a template for other projects that want to produce a package alongside its contents

Signed-off-by: Neil Hickey <nhickey@vmware.com>


packaging/values.yml
```diff
@@ -0,0 +1,6 @@
+#@ load("@ytt:data", "data")
+#@data/values
```
cppforlife: lets use data/values-schema with types of data below.

packaging/metadata.yml
```diff
@@ -0,0 +1,15 @@
+#@ load("@ytt:data", "data")
```
cppforlife: no need for load

packaging/README.md
```diff
@@ -0,0 +1,27 @@
...
+## Configuration
+
+The following configuration values can be set to customize the secretgen-controller installation.
+
+### Secret Gen Controller Configuration
+
+| Value | Required/Optional | Description |
+|-------|-------------------|-------------|
+| `secretgenController.namespace` | Optional | The namespace in which to deploy secretgen-controller.|
+| `secretgenController.createNamespace` | Optional | A boolean that indicates whether to create the namespace specified. Default value is `true`. |
```
cppforlife: no secretgenController prefix anymore. also create_namespace is snake case.

config/values.yml
```diff
@@ -1,8 +1,5 @@
-#@data/values
+#@data/values-schema
 ---
 namespace: secretgen-controller
 create_namespace: true
-
-push_images: false
-image_repo: docker.io/k14s/secretgen-controller-test
-secretgen_controller_version: "develop"
+secretgen_controller_version: "develop"
```
cppforlife: should this be under `dev` since this is only used for building afaik, but hmm it also gets used in the annotation...

neil-hickey: Right, it seems to me like the only way to not have it as a value is to hardcode it in.

config/values.yml
```diff
@@ -1,8 +1,21 @@
...
+#@data/values-schema
 ---
+#@schema/desc "The namespace in which to deploy secretgen-controller"
 namespace: secretgen-controller
+#@schema/desc "Whether to create namespace specified for secretgen-controller"
 create_namespace: true
-
-push_images: false
-image_repo: docker.io/k14s/secretgen-controller-test
 secretgen_controller_version: "develop"
```
neil-hickey: We maybe(?) (actually... very likely) don't want this in our public api for this values file.

This values file will be part of the bundle, but the problem lies https://github.com/vmware-tanzu/carvel-secretgen-controller/pull/65/files#diff-1f0ed1aa0e177024e3bfc3a58979f9d63c23921a83c8e9aa6a6b3b269f728c06L9

We need the version to be a value for our config, or else we would need to hard code it in each time / have automation that edits `config/deployment.yml` each time. This is a stick value, maybe it's fine to expose? But we don't want it to be configurable, or else the user could end up with a difference between the annotation in the deployment, and the actual version of secretgen-controller.

neil-hickey:
Hmm, I need to exclude `config/.imgpkg` from the building and running it seems, the tests are failing on trying to deploy the bundle.yml which it shouldn't.... sigh, weird consequences

Doing something like `ytt -f config/x.yml -f config/y.yml` to get around it seems super ugly... Doesn't look like you can exclude sub paths in ytt..

config/deployment.yml
```diff
@@ -1,13 +1,25 @@
...
+#@ load("@ytt:assert", "assert")
 
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: secretgen-controller
   namespace: #@ data.values.namespace
   annotations:
-    secretgen-controller.carvel.dev/version: #@ data.values.secretgen_controller_version
+    secretgen-controller.carvel.dev/version: #@ data.values.secretgen_controller_version or assert.fail("'secretgen_controller_version' must be set")
```
joe-kimmel-vmw: i tried this `or` thing in the ytt playground 
```

#@ load("@ytt:data", "data")

list_with_a_map: #@ data.values.my_math or 'foo'
```

and just got 
```
- struct has no .my_math field or method (did you mean .my_map?)
    in <toplevel>
      config.yml:6 | list_with_a_map: #@ data.values.my_math or 'foo'
```
      
 are you sure this works?

config/values.yml
joe-kimmel-vmw: it looks like it gets compiled into the binary in the dockerfile:
`28:RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -ldflags="-X 'main.Version=$SGCTRL_VER' -buildid=" -trimpath -o controller ./cmd/controller/...`

and also set as an annotation in the deployment:
```
config/deployment.yml
8:  annotations:
9:    secretgen-controller.carvel.dev/version: #@ data.values.secretgen_controller_version
```


I think we were talking to some other folks about packaging this morning who were saying they basically do two templating passes:  a first one to lock things down before the package is made,  and then a second one that's part of the packaging process.     But i think that amounts to what you describe above as "have automation that edits config/deployment.yml"

packaging/metadata.yml
```diff
@@ -0,0 +1,14 @@
...
+kind: PackageMetadata
+metadata:
+  name: secretgen-controller.carvel.dev
+spec:
+  displayName: "secretgen-controller"
+  longDescription: "Provide CRDs to specify what secrets need to be on cluster (generated or not)."
+  shortDescription: "Secret generation and sharing"
+  providerName: carvel.dev
+  maintainers:
+    - name: carvel.dev
```
joe-kimmel-vmw: to the extent we're trying to make this package look like things TCE wants, the style of "maintainers" that they[ approved previously ](https://github.com/vmware-tanzu/community-edition/pull/3890/files) was `Carvel team (see www.carvel.dev for team members)`

config/deployment.yml
neil-hickey: Can you post a link to the playground example you were trying? It does work for me. Looks like maybe you are trying it on a map, whereas version is just a string here. So perhaps that's why you are getting an error?

https://carvel.dev/ytt/#gist:https://gist.github.com/neil-hickey/3005e22ca03588264c05a4cc42bae47c

packaging/metadata.yml
neil-hickey: oh cool thanks! It's not like that for secretgen. Didn't think to look at KC. 👍  will update

config/deployment.yml
joe-kimmel-vmw: tbh i'm perpetually confused about the playground links,  and the one you sent me doesn't work w/out changing the filename from txt -> yml.  but try this:
```
#@ load("@ytt:assert", "assert")
#@ load("@ytt:data", "data")

#@ p = "123"
#@ f = ""

key1: #@ p or assert.fail("fail")
key2: #@ f or assert.fail("fail")
key3: #@ data.values.foo or assert.fail("fail")
key4: #@ x or assert.fail("fail")
```

in those the p and f work as you want,  but accessing a nonexistent member of the data.values struct triggers a "struct has no ..." before it hits the `or` clause.  (as does asking for a nonexistent variable `x` on the line below)

neil-hickey: Right, if the variable is missing it will fail on the error you pointed out. I figured that was fine, the purpose of the assertion was to ensure you give a non-empty string, not that you give the variable at all. 

I actually got the pattern from the docs:

https://carvel.dev/ytt/docs/v0.41.0/lang-ref-ytt/#assert

`x = data.values.env.mysql_password or assert.fail("missing env.mysql_password")`

config/package-bundle/.imgpkg/bundle.yml
```diff
@@ -0,0 +1,6 @@
+apiVersion: imgpkg.carvel.dev/v1alpha1
```
cppforlife: lets remove this file as its optional

config/package-bundle/.imgpkg/images.yml
```diff
@@ -0,0 +1,14 @@
+---
```
cppforlife: remove this file since it's generated at build time

config/package-bundle/contents/deployment.yml
```diff
@@ -0,0 +1,53 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: secretgen-controller
+  namespace: #@ data.values.namespace
+  annotations:
+    secretgen-controller.carvel.dev/version: v0.8.0
+spec:
+  #@ if data.values.deployment.updateStrategy != "":
+  #@overlay/match missing_ok=True
```
cppforlife: not an overlay, no need overlay/match

config/package-bundle/contents/deployment.yml
```diff
@@ -0,0 +1,53 @@
...
+  strategy:
+    type: #@ data.values.deployment.updateStrategy
+    #@overlay/match missing_ok=True
+    #@ if data.values.deployment.updateStrategy == "RollingUpdate":
+    rollingUpdate:
+      maxUnavailable: #@ data.values.deployment.rollingUpdate.maxUnavailable
+      maxSurge: #@ data.values.deployment.rollingUpdate.maxSurge
+      #@ end
```
cppforlife: nit: for style, unindent 2 spaces

config/package-bundle/contents/deployment.yml
```diff
@@ -0,0 +1,53 @@
...
+          resources:
+            requests:
+              cpu: 120m
+              memory: 100Mi
+          securityContext:
+            runAsUser: 1000
+            runAsGroup: 2000
+      securityContext:
+        fsGroup: 3000
+      #@ if data.values.deployment.nodeSelector != None:
```
cppforlife: can use if/end syntax

config/package-bundle/contents/values.yml
```diff
@@ -0,0 +1,20 @@
+#@data/values-schema
```
cppforlife: should name values-schema files as values-schema.yml @pivotaljohn wdyt?

config/package/package.yml
```diff
@@ -0,0 +1,75 @@
...
+apiVersion: data.packaging.carvel.dev/v1alpha1
+kind: Package
+metadata:
+  name: #@ data.values.name
+spec:
+  refName: #@ data.values.name
+  version: #@ data.values.version
+  releaseNotes: #@ "https://github.com/vmware-tanzu/carvel-secretgen-controller/releases/tag/" + data.values.version
+  valuesSchema:
+    openAPIv3:
```
cppforlife: should be able to generate this via ytt: https://carvel.dev/ytt/docs/v0.41.0/how-to-export-schema/#docs

config/package/package.yml
```diff
@@ -0,0 +1,75 @@
...
+  licenses:
+  - Apache 2.0
+  template:
+    spec:
+      fetch:
+        - imgpkgBundle:
+            image: #@ data.values.image
+      template:
+        - ytt:
+            ignoreUnknownComments: true
```
cppforlife: lets try removing this. (feels like an old copy pasta)

config/release/build.yml
```diff
@@ -1,22 +1,24 @@
...
+      #@ if data.values.dev.push_images:
       #! Always rebuild image
       pull: true
       noCache: true
-      rawOptions: #@ ["--build-arg", "SGCTRL_VER="+data.values.secretgen_controller_version]
+      rawOptions: #@ ["--build-arg", "SGCTRL_VER="+data.values.dev.version]
+      #@ end
 
-#@ if/end data.values.push_images:
+#@ if/end data.values.dev.push_images:
```
cppforlife: for release, we should probably just not even bother making this conditional.

hack/build-release.sh
```diff
@@ -6,7 +6,18 @@ source ./hack/version-util.sh
 
 mkdir -p tmp/
 
-ytt -f config/ -f config-release -v secretgen_controller_version="$(get_sgctrl_ver)" | kbld -f- > ./tmp/release.yml
+export version="$(get_sgctrl_ver)"
+
+yq eval '.metadata.annotations."secretgen-controller.carvel.dev/version" = env(version)' -i "config/package-bundle/contents/deployment.yml"
```
cppforlife: note to future selves: this will make git directory dirty and kbld will record that.

hack/build-release.sh
```diff
@@ -6,7 +6,18 @@ source ./hack/version-util.sh
...
+ytt -f config/package-bundle/contents -f config/release -v dev.version="$version" | kbld --imgpkg-lock-output config/package-bundle/.imgpkg/images.yml -f- > ./tmp/release.yml
+
+imgpkg push -b ghcr.io/vmware-tanzu/carvel-secretgen-controller-package-bundle:"$version" -f config/package-bundle --lock-output ./bundle-image.yml
+
+ytt -f config/package/package.yml -f config/package/values.yml -v version="$version" -v image="$(yq eval '.bundle.image' bundle-image.yml)" > ./tmp/package.yml
+
+cp config/package/metadata.yml ./tmp/metadata.yml
+rm ./bundle-image.yml
 
 shasum -a 256 ./tmp/release*.yml
```
cppforlife: shasum package cr and metadata as well

hack/dev-deploy.sh
```diff
@@ -10,5 +10,7 @@ run_image_start=`cat Dockerfile | grep -n "\- run \-" | cut -d':' -f1`
 tail -n +$run_image_start Dockerfile | \
   sed 's/COPY.*secretgen-controller/COPY controller-linux-amd64 secretgen-controller/' >> Dockerfile.dev
 
-ytt -f config/ -f config-dev-deploy/ | kbld -f- | kapp deploy -a sg -f- -c -y
+yq eval ".dev.rapid_deploy = true" -i config/dev/values.yml
+
+ytt -f config/package-bundle/contents -f config/dev | kbld -f- | kapp deploy -a sg -f- -c -y
```
cppforlife: --data-value-yaml dev.rapid_deploy=true

cppforlife (commented):
went through the PR with @neil-hickey over zoom. we left notes.

config/package-bundle/contents/values.yml
pivotaljohn: sure... or `schema.yml`

But not `values.yml`... that is too often either a Data Values overlay or a Data Values File (i.e. input to `--data-values-file`).

jorgemoralespou:
Is there any test to validate package works and deployment works as expected?

neil-hickey:
> Is there any test to validate package works and deployment works as expected?

There isn't an explicit test for this. There aren't any automated tests for "released artifacts" right now it seems on SG. I will add some more instructions to our release list, and eventually this will be automated.

config/package-bundle/contents/deployment.yml
```diff
@@ -0,0 +1,46 @@
...
+          image: secretgen-controller
+          resources:
+            requests:
+              cpu: 120m
+              memory: 100Mi
+          securityContext:
+            runAsUser: 1000
+            runAsGroup: 2000
+      securityContext:
+        fsGroup: 3000
```
cppforlife: this has been modified on develop to work on openshift.

hack/build-release.sh
```diff
@@ -5,9 +5,27 @@ set -e -x -u
...
 
-shasum -a 256 ./tmp/release*.yml
+yq eval '.metadata.annotations."secretgen-controller.carvel.dev/version" = env(version)' -i "config/package-bundle/contents/deployment.yml"
+
+ytt -f config/package-bundle/contents -f config/release -v dev.version="$version" | kbld --imgpkg-lock-output config/package-bundle/.imgpkg/images.yml -f- > ./tmp/release.yml
+
+imgpkg push -b ghcr.io/vmware-tanzu/carvel-secretgen-controller-package-bundle:"$version" -f config/package-bundle --lock-output ./bundle-image.yml
+
+# generate openapi schema for package
+ytt -f config/package-bundle/contents --data-values-schema-inspect -o openapi-v3 > ./schema-openapi.yml
```
cppforlife: should place tmp files into tmp/

neil-hickey:
@cppforlife I have tested this out on my fork --> https://github.com/neil-hickey/carvel-secretgen-controller/releases/tag/v1.0.0

I have since fixed up the few issues it created, and updated the docs https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/nh-publish-package/docs/dev.md#release to include details on how to release this stuff.

I think for this, the simplest merge will be a squash and merge over a merge commit. There's more value in commiting this as one commit rather than multiple.

config/package-bundle/config/deployment.yml
```diff
@@ -1,13 +1,21 @@
...
+#@ load("@ytt:assert", "assert")
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: secretgen-controller
   namespace: #@ data.values.namespace
   annotations:
-    secretgen-controller.carvel.dev/version: #@ data.values.secretgen_controller_version
+    secretgen-controller.carvel.dev/version: v0.8.0
```
cppforlife: should we set this to `v0.0.0` since it's supposed to get replaced?

config/package-bundle/config/deployment.yml
```diff
@@ -33,3 +41,8 @@ spec:
           capabilities:
             drop:
             - all
+      #@ if/end data.values.deployment.nodeSelector != None:
+      nodeSelector:
+        #@ for/end key in data.values.deployment.nodeSelector:
+        #@yaml/text-templated-strings
+        (@= key @): #@ data.values.deployment.nodeSelector[key]
```
cppforlife: is this just `nodeSelector: #@ data.values.deployment.nodeSelector`

config/package-bundle/config/values-schema.yml
```diff
@@ -0,0 +1,20 @@
...
+#@schema/desc "Configuration for secretgen-controller deployment"
+#@overlay/match-child-defaults missing_ok=True
+deployment:
+  #@schema/desc "Update strategy of deployments, either 'RollingUpdate' or empty (use default strategy)"
+  updateStrategy: "RollingUpdate"
+  rollingUpdate:
+    #@schema/desc "The maxUnavailable of rollingUpdate. Applied only if RollingUpdate is used as updateStrategy"
+    maxUnavailable: 1
+    #@schema/desc "The maxSurge of rollingUpdate. Applied only if RollingUpdate is used as updateStrategy"
+    maxSurge: 0
```
cppforlife: seems like we do not set them in https://github.com/vmware-tanzu/community-edition/blob/main/addons/packages/secretgen-controller/0.9.1/bundle/config/values.yaml#L6-L9. so this would be a change. i think we should not set updateStrategy.

config/package-bundle/config/deployment.yml
neil-hickey: Yeh, actually I am wondering why there was this map situation in TCE / TKG ...

config/package-bundle/config/values-schema.yml
neil-hickey: Good find, I can't recall why I set this one ...

PR #66, opened 2022-05-17
Introduce SecretTemplate API

### LINKED ISSUE #54: introduce SecretTemplate to allow to "shape" secrets
Use cases:
- i want to merge multiple secrets into one
- i want to rename keys within secrets
- i want to hide pieces of the secret so that it can get exported (only exporting ca.crt vs private key)

introduce SecretTemplate CR that is able to read one or more secrets and produce another secret.

for example:
```
---
kind: SecretTemplate
metadata:
  name: combined-secret # will create combined-secret Secret
spec:
  resources:
  - secretRef:
      name: signing-key
  - secretRef:
      name: another
  ytt:
    templateContractV1:
      data.yml: |
        #@ load("@ytt:data", "data")
        metadata:
           annotations:
              foo: bar
        data:
           public.pem: #@ data.values.secrets["signing-key"].data.public
           password-value: #@ data.values.secrets.another.data.value
```

TBD:
- should we allow name glob on resource refs?
- should we allow referencing of configmaps?
- how does baes64 encoding play in here?

[Additional information that will assist in solving the issue.]

---
Vote on this request

Samze:
## Introduce SecretTemplate API

Issue: #54 . 

This PR introduces a new SecretTemplate API. This allows the definition of a number of `InputResources` that will be read and used as part of a `Template` that allows the creation of a new Secret.

The API implemented in this PR:
## API
```yaml
---
apiVersion: secretgen.carvel.dev/v1alpha1
kind: SecretTemplate
metadata:
  name: generated-secret
spec:

  # The Service Account used to read inputResources. If not specified, then only Secrets can be read.
  serviceAccountName: my-resource-reader #optional 
  
   #A list of input resources that are used to construct a new secret. Input Resources can refer to ANY Kubernetes API.
   # If loading more than Secrets types ensure that `.spec.ServiceAccountName` is set to an appropriate value.
   # Input resources are read in the order they are defined. An Input resource's name can be evaluated dynamically from data in a previously evaluated input resource.
  inputResources:
    
    #The name of InputResource. This is used as the identifying name in templating to refer to this Input Resource.
  - name: rds
    ref:
      apiVersion: rds.services.k8s.aws/v1alpha1
      kind: DBInstance
      # The name of the input resource. This field can itself contain JSONPATH syntax to load the name dynamically 
      #from other input resources. For example this field could be set to a static value of "my-secret" or a dynamic valid of 
      # "$(.anotherinputresource.spec.name)".
      name: my-rds-instance

  - name: creds
    ref:
      apiVersion: v1
      kind: Secret
      name: $(.rds.masterPassword.name)

  # A JSONPath based template that can be used to create Secrets.
  template:
     metadata:
       labels:
          key: value
       annotations:
          key: value

     #StringData key and value. Where key is the Secret Key and the value is a JSONPATH syntax surrounded by $( ).
     #All InputResources are available via their identifying name.
     #For example:
     # key1: $(.input1.spec.value1)
     # key2: $(.input2.status.value2)
    stringData:
      type: postgresql
      port: $(.rds.status.endpoint.port)
      database: $(.rds.spec.dbName)
      host: $(.rds.status.endpoint.address)
      username: $(.rds.spec.masterUsername)
     # Data key and value. Where key is the Secret Key and the value is a jsonpath surrounded by $( ). The fetched data MUST be base64 encoded.
     # All InputResources are available via their identifying name.
     # For example:
     # key1: $(.secretinput1.data.value1)
     # key2: $(.secretinput2.data.value2)
    data:
      password: $(.creds.data.password)
status: 
  secret:
    name: generated-secret
  conditions:
  - lastTransitionTime: 2019-10-22T16:31:24Z
    status: True
    type: ReconcileSucceeded
``` 

## Notes
1. Only implemented basic JSONPath templating and does not support a `ytt` as defined in #54 yet. For now we just support `.spec.template` which is a required field. Once `.spec.ytt` is added we can make `template` optional and add validation that ensures either is set.
1. We utilise the Kubernetes library https://github.com/kubernetes/client-go/blob/master/util/jsonpath/jsonpath.go for JSONParsing - this supports a subset of jsonpath. 
1. We currently delete the generated Secret on any reconcile failure. This is a super cautious approach. This is due to ensuring that if an InputResource Secret is deleted and so we fail to resolve an input resource that we cleanup the generated secret.
1. We currently "adopt and overwrite" any existing secret with the same name. This aligns us with other secret-gen resources but may be potentially dangerous.
1. In the original design in #54 , we had the additional conditions for `InputResourcesFound`, `TemplateSucceeded` and `SecretCreated`. We decided these didn't fit the style of the current resources and we have no direct usecases for them. We can add them in the future if needed.
1. There was discussion in #54 around whether this resource should implement `ProvisionedService` as defined https://github.com/k8s-service-bindings/spec. We have kept this out of this PR for now. We may create a follow up issue if necessary. 
1. Similar to above, there was some in #54 around immutability of secrets and whether we should delete/recreate. We have kept this out of this PR for now.
1. We added permissions for secret-gen controller to read serviceaccounts so that it can read the SA token to create a k8s client from `.spec.serviceAccountName`.
1. For now we just requeue the SecretTemplate reconciler to pickup changes to InputResources, this is similar to kapp controller.  It is currently set to 30seconds. We may want to make this configurable in in the future, or better yet add something smart like dynamic watching of InputResources.

## Outstanding items:
- [ ] `/docs` folder. We need to agree on a way forward with the existing secretgen1 `SecretTemplate` to add these. See question below.

## Questions
1. There is an existing `SecretTemplate` [struct](https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/pkg/apis/secretgen/v1alpha1/secret_template.go). We have not touched it so far as the struct is embedded into other types (e.g. Password) but there is heavy overlap here as well as a shared name. It also appears to be documented here https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/docs/secret-template.md. Any thoughts on what we should do here with the overlap? 
1. A Secrets `Type` field is immutable and so if someone updates the SecretTemplate `.spec.template.type` we cannot simply update the Secret. We currently followed the the behaviour of existing resources like `Password` which is to ignore the updated value. We could alternatively delete and recreate the secret. Thoughts?

closes #54

Cheers,

@gmrodgers @Samze


pkg/apis/secretgen2/v1alpha1/secret_template.go
```diff
@@ -0,0 +1,105 @@
...
+	InputResources []InputResource `json:"inputResources"`
+
+	//A JSONPath based template that can be used to create Secrets.
+	// +optional
+	JSONPathTemplate JSONPathTemplate `json:"template,omitempty"`
+
+	//The Service Account used to read inputResources. If not specified, then the Service Account of
+	//secretgen controller is used.
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
```
st3v, line 56: I guess you're optimizing for usability by making this optional but this smells like it has the potential for privilege escalation, in particular if this API will ever evolve to support input resources from multiple namespaces. But even for the single namespace case this could be problematic. What are the RBAC permissions of the service account secretgen controller uses today? What will they be in the future? If they are (or ever will be) for anything else than secrets in the same namespace as the template itself, there's a potential attack vector.

Samze: Yeah this was for convenience as suggested in https://github.com/vmware-tanzu/carvel-secretgen-controller/issues/54 . To guard against this we can ensure that if a ServiceAccount is not specified then we only allow loading secrets in the same namespace. In the future if we add namespaces support we must ensure that any cross namespace loading of any resource requires a ServiceAccount.

st3v: How do you feel about my concern w/r to the single namespace use case? Even if the SA for the secretgen controller is locked down to only `secrets` right now, how do we make sure it stays that way and we don't accidentally introduce the potential for privilege escalation as a side effect of an unrelated change elsewhere in this repo?

st3v: If you want to get to the level of convenience you're after without coupling this to the RBAC of the controller SA, would it make sense to have a separate SA that is deployed alongside secretgen controller and used for the sole purpose to act as the default for `secrettemplates`? That SA could be named accordingly and we could bind it to a dedicated clusterrole that is **and forever will be** locked down to `secrets`.

Samze: > How do you feel about my concern w/r to the single namespace use case? Even if the SA for the secretgen controller is locked down to only secrets right now, how do we make sure it stays that way and we don't accidentally introduce the potential for privilege escalation as a side effect of an unrelated change elsewhere in this repo?

I'm suggesting we still use the secret-gen controller client but programmatically ensure we only ever load Secrets from the same namespace as SecretTemplate and error if any other type of inputresource/namespace is specified. That way we guard against your concern that secret-gen controller may gain permission in the future.

Samze: Caught up with @st3v about this offline. We think there are three options.
1. Require `.spec.serviceAccountName` and always use that to load input resources. 
1. Keep `.spec.serviceAccountName` optional but when not specified our reconciler programmatically only allows the loading of Secrets. [Like this](https://github.com/vmware-tanzu/carvel-secretgen-controller/pull/66/files#diff-99938f4a9e278a1f7931cbf6e46448b7780c38f1e65d5045e213ad99c303b6e2R223).
1. At installation time create a new dedicated ServiceAccount and corresponding ClusterRole and ClusterRoleBinding that only allows to the reading of secrets. We use this as the identity when a ServiceAccount is not specified.

Option 1 gives a consistent UX but for the simple case of combining secrets requires more setup (SA/Role/RoleBinding). It also has the benefit of allowing us to always rely on the k8s RBAC.

Option 2 gives a simpler UX for the combing secret case, since no SA is needed, however from a UX perspective the API is more complicated - so is maybe more confusing. Additionally we must enforce that only Secret can be loaded in our reconciler, which may be more error prone.

Option 3 is the same UX as Option 2. It does allow us to rely more heavily on RBAC but means we must deploy additional resources as part of the secret-gen installation.

@cppforlife I think you originally suggested this optionality in #54:
> for general secretgen users majority of use cases would be covered by "without service account". for more advanced cases you have, we would definitely need "service account" support. i think we should follow the same methodology as with jsonpath (+ ytt for advanced). no-service-account for general use, with-service-account for advanced cases.

Do you have any insights on how common the "secrets are the only input resources" usecase is and do you have any opinions here on the security model?

pkg/apis/secretgen2/v1alpha1/secret_template.go
```diff
@@ -0,0 +1,129 @@
...
+
+// SecretTemplateSpec contains spec information
+type SecretTemplateSpec struct {
+	//A list of input resources that are used to construct a new secret. Input Resources can refer to ANY Kubernetes API.
+	//If loading more than Secrets types ensure that `.spec.ServiceAccountName` is set to an appropriate value.
+	//Input resources are read in the order they are defined. An Input resource's name can be evaluated dynamically from data in a previously evaluated input resource.
+	InputResources []InputResource `json:"inputResources"`
+
+	//A JSONPath based template that can be used to create Secrets.
+	JSONPathTemplate JSONPathTemplate `json:"template"`
```
cppforlife: lets turns this into a pointer in preparation for other templates (e.g. ytt)

pkg/apis/secretgen2/v1alpha1/secret_template.go
```diff
@@ -0,0 +1,129 @@
...
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds
+	// +optional
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	Items []SecretTemplate `json:"items"`
+}
+
+// SecretTemplateSpec contains spec information
+type SecretTemplateSpec struct {
+	//A list of input resources that are used to construct a new secret. Input Resources can refer to ANY Kubernetes API.
```
cppforlife: nit: can we format all comments so that they have a space after `//`

pkg/apis/secretgen2/v1alpha1/secret_template.go
cppforlife: im in favor of option 2.

>  Additionally we must enforce that only Secret can be loaded in our reconciler, which may be more error prone.

i think we should be able to reach a confident level here via good code structure and tests.

pkg/apis/secretgen2/v1alpha1/secret_template.go
```diff
@@ -0,0 +1,129 @@
...
+	sgv1alpha1.GenericStatus `json:",inline"`
+	// +optional
+	ObservedSecretResourceVersion string `json:"observedSecretResourceVersion,omitempty"`
+}
+
+// Validate runs validation on the SecretTemplate
+func (e SecretTemplate) Validate() error {
+	var errs []error
+
+	//TODO
```
cppforlife: anything to do?

pkg/generator/secret_template_reconciler.go
```diff
@@ -0,0 +1,272 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// Matches default sync in kapp-controller
+	// See: https://github.com/vmware-tanzu/carvel-kapp-controller/blob/develop/pkg/app/reconcile_timer.go
```
cppforlife: dont think we should call this out here. comment almost seems to imply some form of relationship but there is not, right?

pkg/generator/secret_template_reconciler.go
```diff
@@ -0,0 +1,272 @@
...
+	defer r.updateStatus(ctx, &secretTemplate)
+
+	res, err := r.reconcile(ctx, &secretTemplate)
+	//TODO is this overly defensive?
+	if err != nil {
+		secretTemplate.Status.Secret.Name = ""
+		if deleteErr := r.deleteChildSecret(ctx, &secretTemplate); deleteErr != nil {
+			return status.WithReconcileCompleted(res, deleteErr)
+		}
+	}
```
cppforlife: i dont think we should be removing existing secret on failure since that would cause flip flopping behaviour, but i do think we should clearly indicate in the status that things have gone wrong (which i think it does)

pkg/generator/secret_template_reconciler.go
```diff
@@ -0,0 +1,272 @@
...
+	secret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretTemplate.GetName(),
+			Namespace: secretTemplate.GetNamespace(),
+		},
+	}
+
+	if _, err = controllerutil.CreateOrUpdate(ctx, r.client, &secret, func() error {
+		secret.ObjectMeta.Labels = secretTemplate.Spec.JSONPathTemplate.Metadata.Labels
+		secret.ObjectMeta.Annotations = secretTemplate.Spec.JSONPathTemplate.Metadata.Annotations
```
cppforlife: we dont allow json path templating in labels and annotations?

pkg/generator/secret_template_reconciler.go
```diff
@@ -0,0 +1,272 @@
...
+		secret.StringData = secretStringData
+		secret.Data = secretData
+
+		// Secret Type is immutable, so cannot be updated. TODO what to do here?
```
cppforlife, line 119: great question, delete/create?

pkg/generator/secret_template_reconciler.go
```diff
@@ -0,0 +1,272 @@
...
+
+		return controllerutil.SetControllerReference(secretTemplate, &secret, scheme.Scheme)
+	}); err != nil {
+		return reconcile.Result{}, fmt.Errorf("creating/updating secret: %w", err)
+	}
+
+	secretTemplate.Status.Secret.Name = secret.Name
+
+	return reconcile.Result{
+		RequeueAfter: syncPeriod,
```
cppforlife, line 130: hmm, do we have any other ideas for the future how to make this more performant?

at least for when inputs are all secrets, can we only trigger reconciliation only based on watches?
for case where there are non-secret inputs, im not even sure how to do this best...

pkg/generator/service_account_loader.go
```diff
@@ -0,0 +1,102 @@
...
+	return &rest.Config{
+		Host: cfg.Host,
+		TLSClientConfig: rest.TLSClientConfig{
+			CAData: cert,
+		},
+		BearerToken: string(token),
+	}, nil
+}
+
+func (s *ServiceAccountLoader) serviceAccountCredentials(ctx context.Context, name, namespace string) ([]byte, []byte, error) {
```
cppforlife: we ll want to change this to use TokenRequest API (k8s 1.24 break this implementation):

- https://github.com/vmware-tanzu/carvel-kapp-controller/issues/687
- beginnings of kc change: https://github.com/vmware-tanzu/carvel-kapp-controller/pull/695

test/e2e/secret_template_test.go
```diff
@@ -0,0 +1,391 @@
...
+			Message: "cannot fetch input resource secret1: secrets \"secret1\" not found",
+		})
+
+		var secretTemplate sg2v1alpha1.SecretTemplate
+		err := yaml.Unmarshal([]byte(out), &secretTemplate)
+		if err != nil {
+			t.Fatalf("Failed to unmarshal: %s", err)
+		}
+
+		if !reflect.DeepEqual(secretTemplate.Status.Secret.Name, "") {
```
cppforlife: please switch to assert.X/require.X (`github.com/stretchr/testify/assert`) -- we've been trying to only use that in our new tests (havent updated all existing tests though).

pkg/generator/service_account_loader.go
cppforlife: cc @joe-kimmel-vmw @neil-hickey

pkg/generator/secret_template_reconciler.go
Samze: This is totally possible, do you have any specific usecases in mind? We're open to adding this.

In general, perhaps all values within `JSONPathTemplate` allow templating. Practically this would mean adding templating to `type`, `labels` and `annotations.

pkg/generator/secret_template_reconciler.go
Samze: Are you suggesting deleting/creating in all cases or just when type is updated?

pkg/apis/secretgen2/v1alpha1/secret_template.go
Samze: Fixed

pkg/apis/secretgen2/v1alpha1/secret_template.go
Samze: Removed, I think we may want to add this back once we have either ytt / template to validate that one is present.

pkg/generator/secret_template_reconciler.go
Samze: Removed

pkg/generator/secret_template_reconciler.go
cppforlife: > This is totally possible, do you have any specific usecases in mind?

if for whatever reason annotation need to be carried over from some input resource. but generally, yeah, seems a bit awkward if not all fields go through same "mutation"

pkg/generator/secret_template_reconciler.go
cppforlife: only when type is updated since it not possible to reconcile.

pkg/generator/service_account_loader.go
joe-kimmel-vmw: after https://github.com/vmware-tanzu/carvel-secretgen-controller/pull/67 is merged you can pull-rebase to get some test failures when you access SA creds the 'old' way.

pkg/generator/secret_template_reconciler.go
gmrodgers: Should now have the ability to have templating in `type`, `labels` and `annotations`.

pkg/generator/service_account_loader.go
gmrodgers: Looks like there's no issue for secretgen capturing what you did in #67 @joe-kimmel-vmw.  For now I think we'll copy over the satoken package and utilise it in our service_account_loader as part of this PR!

pkg/generator/secret_template_reconciler.go
gmrodgers: This is probably fine for us but there is a bit of a weird unexpected behaviour change versus if the type is changed vs when anything else in the template is changed.  The same would go if immutability was set on the Secret.  If we're fine with just documenting that difference then maybe we can go ahead.

pkg/apis/secretgen2/v1alpha1/secret_template.go
gmrodgers: Just to check this, were you suggesting just to make this a pointer, or to make it an optional field (with +optional, omitempty etc) as well?

pkg/generator/secret_template_reconciler.go
Samze: Yeah this is an interesting problem and yes ideally we would be notified of changes to input resources through watches.

For arbitrary input resources this is made hard by the fact we use a separate Service Account for each SecretTemplate. This means that we cannot simply setup a single informer & watch for a GVK (since we have multiple identities and the watch call must use one identity) and on changes map to a corresponding SecretTemplate. So the options I see are:
1. Rely on period resync (as implemented in this PR)
1. Create a new watch **per** GVK (or specific input resource) **per** SecretTemplate Service Account.

For option 2, given that would mean an explosion of watches to the API Server I think that may be even less scalable than the period resync. 

Have you explored this problem as part of https://github.com/vmware-tanzu/carvel-kapp-controller/ ? Seems very similar, user defined ServiceAccounts that access arbitrary Resources.

For Secrets it's a different matter because we use the same controller service account. This means we just need to watch for secrets and having a mapping function to the SecretTemplate.

pkg/generator/secret_template_reconciler.go
Samze: Our main concern here was the scenario of when an input resource, such as a secret, being deleted and so fails to resolve and reconcile errors. We thought it was risky to keep the generated Secret around with potentially sensitive data that the user had thought they had deleted from the cluster. Equally this could be an issue if a field is deleted from a secret.. Do you have thoughts on these scenario? 

So the options I see are:
1. Be defensive and accept that disappearing/reappearing secrets may cause other consumers issues.
1. We could be smarter on the cases when deleting the child secret. For example, delete the secret if some information is missing (e.g. inputresource not found, or field not resolvable).
1. Not remove the Secret on Reconcile failure and rely on the status, potentially leaving data around.

pkg/generator/secret_template_reconciler.go
Samze: I spiked out just watching secrets here, https://github.com/Samze/carvel-secretgen-controller/pull/1/files we can do a follow up PR with these sort of changes.

pkg/generator/secret_template_reconciler.go
```diff
@@ -0,0 +1,322 @@
...
+
+	status := &reconciler.Status{
+		S:          secretTemplate.Status.GenericStatus,
+		UpdateFunc: func(st sgv1alpha1.GenericStatus) { secretTemplate.Status.GenericStatus = st },
+	}
+
+	status.SetReconciling(secretTemplate.ObjectMeta)
+	defer r.updateStatus(ctx, &secretTemplate)
+
+	res, err := r.reconcile(ctx, &secretTemplate)
```
Samze: A potential issue we have here is that we return errors on any failure. This means that we are at the mercy of controller-runtimes exponential backoff - which goes up to [16minutes](https://github.com/kubernetes/client-go/blob/master/util/workqueue/default_rate_limiters.go#L41). For cases when an InputResource or a field is missing and appears eventually, this may not be a desirable amount of time to wait. 

When not using a custom ServiceAccount and InputResources are just secrets this isn't so much of an issue because we can rely on watch, see [thread](https://github.com/vmware-tanzu/carvel-secretgen-controller/pull/66#discussion_r878121055).

For custom ServiceAccount InputResources this becomes more of an issue.

pkg/generator/secret_template_reconciler.go
cppforlife: yeah, im also on the fence about this...

i fear deletion behaviour will cause instability and outages. (e.g. im changing my SecretTemplate to ref secret2 instead of secret1, while at the same time deleting secret1. if it happens that secret1 is deleted first, my result secret is gone which may be is being consumed by some app that now has an outage. when secret2 is brought back in, result secret is "fixed"; or if you have screwed up your templating your secret gets deleted and again your apps potentially incur outage).

i think generally i would side with system should not change created resources if it cannot construct them successfully (in our case due to templating errors).

> We thought it was risky to keep the generated Secret around with potentially sensitive data that the user had thought they had deleted from the cluster.

once the data touches the cluster all bets are off in my opinion. removing it will not "undo the damage" -- if something is truly sensitive and it's leaked, it should be invalidated with a system that cares about that credential.

is there some precedent in k8s core or other projects?

pkg/generator/service_account_loader.go
gmrodgers: So the latest commits now include a version of the `satoken` package.  Initially it was failing in the e2e tests due to the serviceaccount being deleted and then recreated but the cache could not distinguish here. There are two things I can see us doing:
1. Use a cache key that is universal across all serviceaccounts.  We could get the SA and then use its UID perhaps
2. Use the TokenReview API which takes it into account.

Currently the implementation goes for 2) but let us know if yous have a preference.

gmrodgers: Also note that we believe Kapp controller will have the same issue so the fix could be ported over.

gmrodgers: I have created an issue in kapp with steps that I believe will trigger the same issue https://github.com/vmware-tanzu/carvel-kapp-controller/issues/701#issue-1248798176

voor:
Does the example support embedding in deeper levels, like:

```
  # A JSONPath based template that can be used to create Secrets.
  template:
     metadata:
       labels:
          key: value
       annotations:
          key: value

     #StringData key and value. Where key is the Secret Key and the value is a JSONPATH syntax surrounded by $( ).
     #All InputResources are available via their identifying name.
     #For example:
     # key1: $(.input1.spec.value1)
     # key2: $(.input2.status.value2)
    stringData:
      fieldname: "hardcoded-string: $(.input1.spec.value1)"
      longerexample.com: "hardcoded-field: $(.input1.spec.value1)\nmorewords: $(.input2.status.value2)\nbob: hello\n: alice: good morning"

```

Samze:
> Does the example support embedding in deeper levels, like:
> 
> ```
>   # A JSONPath based template that can be used to create Secrets.
>   template:
>      metadata:
>        labels:
>           key: value
>        annotations:
>           key: value
> 
>      #StringData key and value. Where key is the Secret Key and the value is a JSONPATH syntax surrounded by $( ).
>      #All InputResources are available via their identifying name.
>      #For example:
>      # key1: $(.input1.spec.value1)
>      # key2: $(.input2.status.value2)
>     stringData:
>       fieldname: "hardcoded-string: $(.input1.spec.value1)"
>       longerexample.com: "hardcoded-field: $(.input1.spec.value1)\nmorewords: $(.input2.status.value2)\nbob: hello\n: alice: good morning"
> ```

Yeah @voor it does support multiple jsonpath expressions for a single field. See this part of the [non-secret-input example](https://github.com/vmware-tanzu/carvel-secretgen-controller/pull/66/files#diff-f8ae714f0982a3715a91892741459e9081d3ac2877e827d209135898f131fbb4R44). 

Taking that example 's input resources and your stringData key `longerexample.com: "hardcoded-field: $(.config.data.key1)\nmorewords: $(.config.data.key2)\nbob: hello\nalice: good morning"`

Produces this:
```sh
k secretdata non-secret-inputs -n default
default:
  non-secret-inputs:
    longerexample.com: |-
      hardcoded-field: val1
      morewords: val2
      bob: hello
      alice: good morning
```

cppforlife:
@voor we ll try to follow up with ytt templating as well. for anything aside from simple mapping i would not recommend jsonpath (escaping etc becomes a problem quickly).

pkg/generator/secret_template_reconciler.go
cppforlife: i dont think this needs to be a blocker for this PR, but definitely feels a bit weird having this special case when secret cannot be updated (i blame k8s core api for this though).

pkg/apis/secretgen2/v1alpha1/secret_template.go
cppforlife: both, so that we can check within reconciler code wehther JSONPathTemplate is given vs YttTemplate.

pkg/generator/secret_template_reconciler.go
Samze: @cppforlife no prior art that I'm aware of. Happy to go with option 3 for this PR. We can always improve this in the future.

pkg/generator/secret_template_reconciler.go
Samze: I think we can address this in a subsequent PR.

Samze:
Comments and feedback addressed, so the first pass of this is ready to be merged. Additional Issues/PRs include:
- Using Watch rather than periodic reconcile for when SecretTemplate is used for Secrets only (does not specify a Service account)
- Align on the implementation for token refreshing for k8s 1.24 with https://github.com/vmware-tanzu/carvel-kapp-controller/issues/701
- Deleting/Creating the secret when `type` changes due to its immutability
- Reduce reliance on exponential backoff to avoid long reconcilation times when input resources eventually contain the required data.
- Add additional documentation

voor:
image

PR #69, opened 2022-05-27
SecretTemplate watches for Secret InputResources

### LINKED ISSUE #68: SecretTemplate should rely on Watch when a Service Account is not specified

#66 Introduced the SecretTemplate API. This allows you to define arbitrary input resources that are used to generate a new Kubernetes Secret based on templating. 

For a first pass, to handle updates to these input resources we decided to simply periodically re-sync all input resources. However this isn't very performant or scalable.

Instead of a periodic resync we should rely on watches. However SecretTemplate allows the user to provide a Service Account, this means we cannot simply have a single watch between clients and share informers. Each watch call is made for a particular identity. So any sharing mechanism would need some serious thought.

However, for cases when a Service Account is not present, SecretTemplate can only read Secrets as Input Resources. Therefore we can use watch for the non-service account secrets case.

Therefore we should do the following
- When a Service Account is not specified (Secrets only), we rely on Watches to be informed on updates.
- When a Service Account is specified, we continue to periodically resync.

Addressing the general problem can be a future enhancement.

Addtional context in [this PR thread](https://github.com/vmware-tanzu/carvel-secretgen-controller/pull/66/files#r878121055)

---
Vote on this request

Samze:
This introduces a watch mechanism for SecretTemplate when no service
account is specified, so for this case we can remove the periodic
re-sync - improving performance.

We still need a periodic resync for non-secret resources because we
use a user configurable service account for these resources.
Therefore we cannot use a single watch or shared informer and setting
up watches per secret template service account per GVR is too expensive.

Implementation notes:
- As mentioned only support watch when Service Account is not specified
- Rely on an internal tracker that can be referred when a secret event
in generated. This tracker allows a SecretTemplate to track N Secrets.
And for a given Secret, allows you to receive all interested
SecretTemplates.
- When a SecretTemplate is deleted we use 404 to clear the tracking
 information. Even if we don't receive a reconcile immediately, if an
    input resource changes, we will receive a notfication for the missing
    SecretTemplate and then clear tracking that way.
- Tracker currently doesn't understand GVKs, right now this is not needed.

closes #68

### DIFF
--- a/pkg/generator/secret_template_reconciler.go
+++ b/pkg/generator/secret_template_reconciler.go
@@ -204,6 +204,16 @@ func (r *SecretTemplateReconciler) resolveInputResources(ctx context.Context, se
 
 	resolvedInputResourceKeys := []types.NamespacedName{}
 	resolvedInputResources := map[string]interface{}{}
+
+	// Ensure we track any resources we have resolved, even if we failed to resolve them all.
+	defer func() {
+		if shouldTrackInputResources(secretTemplate) {
+			//Untrack everything first incase input resources have changed.
+			r.secretTracker.UntrackAll(secretTemplateKey)
+			r.secretTracker.Track(secretTemplateKey, resolvedInputResourceKeys...)
+		}
+	}()
+
 	for _, inputResource := range secretTemplate.Spec.InputResources {
 		// Ensure we only load Secrets if using the default Client.
 		if secretTemplate.Spec.ServiceAccountName == "" && (inputResource.Ref.Kind != "Secret" || inputResource.Ref.APIVersion != "v1") {
@@ -223,12 +233,6 @@ func (r *SecretTemplateReconciler) resolveInputResources(ctx context.Context, se
 
 		resolvedInputResources[inputResource.Name] = unstructuredResource.UnstructuredContent()
 		resolvedInputResourceKeys = append(resolvedInputResourceKeys, key)
-	}
-
-	if shouldTrackInputResources(secretTemplate) {
-		//Untrack everything first incase input resources have changed.
-		r.secretTracker.UntrackAll(secretTemplateKey)
-		r.secretTracker.Track(secretTemplateKey, resolvedInputResourceKeys...)
 	}
 
 	return resolvedInputResources, nil
--- a/pkg/generator/secret_template_reconciler.go
+++ b/pkg/generator/secret_template_reconciler.go
@@ -38,18 +38,26 @@ type ClientLoader interface {
 	Client(ctx context.Context, saName, saNamespace string) (client.Client, error)
 }
 
+// Tracker allows a tracking resource to track multiple other resources
+type Tracker interface {
+	Track(tracking types.NamespacedName, tracked ...types.NamespacedName)
+	UntrackAll(tracking types.NamespacedName)
+	GetTracking(tracked types.NamespacedName) []types.NamespacedName
+}
+
 // SecretTemplateReconciler watches for SecretTemplate Resources and generates a new secret from a set of input resources.
 type SecretTemplateReconciler struct {
-	client   client.Client
-	saLoader ClientLoader
-	log      logr.Logger
+	client        client.Client
+	saLoader      ClientLoader
+	secretTracker Tracker
+	log           logr.Logger
 }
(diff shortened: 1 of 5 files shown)


pkg/generator/secret_template_reconciler.go
```diff
@@ -75,7 +103,21 @@ func (r *SecretTemplateReconciler) Reconcile(ctx context.Context, request reconc
 		return reconcile.Result{}, err
 	}
 
-	if secretTemplate.DeletionTimestamp != nil {
+	if secretTemplate.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(&secretTemplate, finalizerName) {
```
gmrodgers: *suggestion*: AddFinalizer will only add if it doesn't already exists so we can remove the contains check.

pkg/generator/secret_template_reconciler.go
```diff
@@ -75,7 +103,21 @@ func (r *SecretTemplateReconciler) Reconcile(ctx context.Context, request reconc
...
+			controllerutil.AddFinalizer(&secretTemplate, finalizerName)
+			if err := r.client.Update(ctx, &secretTemplate); err != nil {
```
gmrodgers: *thought*: Anyway we can do a similar `defer` update statement here or is it messy

pkg/generator/secret_template_reconciler.go
Samze: I think we still need the contains check because we only want to update the resource if the finalizer has been added.

pkg/generator/secret_template_reconciler.go
Samze: Mmm maybe, I'm not here it makes so much sense because we only want to call update in certain circumstances. (finalizer added or removed).

pkg/generator/secret_template_reconciler.go
```diff
@@ -202,6 +255,13 @@ func resolveInputResource(ref sg2v1alpha1.InputResourceRef, namespace string, in
 	return toUnstructured(ref.APIVersion, ref.Kind, namespace, resolvedName.String())
 }
 
+// Returns whether we should track the resources contained in a SecretTemplate.
+// We only track resources when a ServiceAccountName has not been specified. This implicitly means
+// we only track Secret resources.
+func isTrackingInputResources(s *sg2v1alpha1.SecretTemplate) bool {
```
gmrodgers: *suggestion non-blocking*: could this be something like `isTrackingRequired`, I was a bit confused at first

Samze: I wonder if the `is` is slightly confusing here as `*sg2v1alpha1.SecretTemplate` itself isn't tracking resources. Maybe `shouldTrack` or `shouldTrackInputResources` is better?

pkg/generator/secret_template_reconciler.go
cppforlife: any way we can avoid finalizer here? without finalizer i guess we could lose "events" for deletion of secrets/secrettemplate...

Samze: This was really to ensure we tidy up the tracking resources when a SecretTemplate is Deleted.

Thinking about this though I guess what we could do is if we get a Reconcile event that results in a 404 for SecretTemplate, we can clear that Templates tracking resources.

pkg/generator/secret_template_reconciler.go
gmrodgers: `shouldTrackInputResources` works for me 👍

pkg/generator/secret_template_reconciler.go
gmrodgers: Makes sense, not blocking for me.

pkg/generator/secret_template_reconciler.go
Samze: Changed this to remove the finalizer and instead clear tracking information when the SecretTemplate is not found. This will mean even if we miss the deletion event for SecretTemplate, any subsequent change to an input resource will result in a 404 and we will clear tracking data.

cppforlife: nice!

PR #73, opened 2022-06-08
Reduce SecretTemplates max exponential backoff

### LINKED ISSUE #71: Failing SecretTemplate with missing InputResources takes a long time to re-reconcile
**What steps did you take:**

1. Create a SecretTemplate with non-secret InputResources that do not exist yet
2. Wait 5 minutes
3. Create the InputResources

**What happened:**
You must wait numerous minutes for reconcilation to occur and for SecretTemplate to show `Ready=true`.

**What did you expect:**
SecretTemplate to re-reconcile faster.

As discussed in https://github.com/vmware-tanzu/carvel-secretgen-controller/issues/68 this is not a problem for when a Service Account is not specified and InputResources are secrets, we rely on watch mechanisms to notice changes. However we do not use Watch for non-secret inputs (See above issue for details why)

When SecretTemplate specifies a ServiceAccount and successfully reconciles we currently periodically requeue to detect changes to those InputResources. However, when reconciliation returns an error, we are currently relying on controller-runtime's exponential backoff mechanism. This means it will maximally only requeue every [16minutes](https://github.com/kubernetes/client-go/blob/911ef75/util/workqueue/default_rate_limiters.go#L39).

For error conditions that may eventually resolve without making changes to SecretTemplate, we should requeue rather than rely on the backoff. The conditions where this is appropriate are:
- When an input resource cannot be fetched
- When templating fails because a resolved input resource does not contain the necessary fields

(Note: This was raised in the original PR here https://github.com/vmware-tanzu/carvel-secretgen-controller/pull/66#issuecomment-1139022727)

---
Vote on this request

We are also happy to receive and review Pull Requests if you want to help working on this issue.

Samze:
* The default controller exponential backoff is 1000*time.Second (16.6
minutes)
* SecretTemplate has input resources that may appear or change that may
result in a successfuly reconcile. We therefore do not want to have to
wait up until 16 minutes for this. Instead configure a max exponential
backoff of 2 minutes.

closes #71

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -9,6 +9,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/go-logr/logr"
 	sgv1alpha1 "github.com/vmware-tanzu/carvel-secretgen-controller/pkg/apis/secretgen/v1alpha1"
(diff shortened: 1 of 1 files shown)


Samze:
CI looks to be failing on an unrelated flake:
```
kapp: Error: Expected all resources to be gone, but found: secret/rsa-key1 (v1) namespace: default
```

cppforlife:
looking into flake separately.

PR #83, opened 2022-07-01
Add automatic release process

neil-hickey:
- Also add a post draft release test for new PackageCR

Validated on my fork: see action run here https://github.com/neil-hickey/carvel-secretgen-controller/actions/runs/2598559248

and https://github.com/neil-hickey/carvel-secretgen-controller/releases/tag/v1.3.12 is the actual release generated by the script tested out on the fork also.

Signed-off-by: Neil Hickey <nhickey@vmware.com>


docs/dev.md
```diff
@@ -21,75 +21,40 @@ the loop of iterating and re-deploying small changes, you will definitely prefer
...
-`kapp deploy -a secretgen-package -f package.yml -f metadata.yml -f pkgi.yml`
-Ensure the deployment and build worked by running the e2e tests:
-you may need to `export SECRETGEN_E2E_NAMESPACE=secretgen-test`
-then `./hack/test-e2e.sh`.
-
-After verifying, commit: `git add releases; git commit -m "release 1.2.3"`
-and push: `git push origin HEAD --tags`
+Push the tag to GitHub.
+```
+git push --tags
```
praveenrewar: Should we suggest `git push origin v1.2.3` instead? (`git push --tags` would push all the tags present in local which might not be desired)

.github/workflows/release.yml
```diff
@@ -0,0 +1,173 @@
...
+        with:
+          name: ${{ github.ref_name }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          body_path: ./tmp/checksums.txt
+          files: |
+            ./tmp/release.yml
+            ./tmp/package.yml
+            ./tmp/metadata.yml
+          draft: true
+          prerelease: true
```
praveenrewar: I think `prerelease` is not required (Unless we want to mark the releases as prereleases)

.github/workflows/release.yml
```diff
@@ -0,0 +1,173 @@
...
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: |
+          set -e -x
+          cat ./tmp/checksums.txt
+          diff ./tmp/checksums.txt <(cat <<EOF
+          ${{steps.get-checksums-from-draft-release.outputs.result}}
+          EOF
+          )
+  validate-package-cr:
```
praveenrewar: Would it make sense to run the e2e tests before creating the draft release as a separate step? This would remove some redundant steps like checkout, setup Go, installing carvel tools and setting up minikube and a draft release wouldn't be created if the e2e tests are failing. 
(Although I guess the the reason for having a separate job might be to validate the draft release, but it also seems like testing the action that creates the draft release)

docs/dev.md
neil-hickey: sure thing, just took this verbatim from the kapp-controller doc. I like this suggestion though 👍

.github/workflows/release.yml
neil-hickey: Right, this is the "hey we successfully built a release, and it's in draft - we ought to validate that the PackageCR that was generated and the live imgpkg bundle works"

I am not really sure if adding this beforehand adds much, other than maybe just not having a draft release created, but it's so easily deletable / overridable. Didn't feel worth it. Some of the steps are duplicated, but that's just because we need a minikube etc to really validate the PackageCR works on a live env.

But maybe I am missing a concern of yours

.github/workflows/release.yml
neil-hickey: yeh, seems redundant. Another port from KC.

.github/workflows/release.yml
praveenrewar: I don't really have any strong opinions, I just thought that having a separate step in the same job would be beneficial in 2 ways. 
- Draft release is not created if there's an issue with the package.yml generated.
- Having a single job would eliminate a few redundant steps which would slightly decrease the workflow run time.

But just like you have mentioned, these are not very significant advantages, so we can go ahead with two separate jobs :)

neil-hickey: ah i see. Yanno maybe it's fine to merge em, while I was testing it was nice to only re-run the failed job. I am starting to lean to the append this testing to the end of the other job.

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18
+      - name: Install Carvel Tools
+        uses: vmware-tanzu/carvel-setup-action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
cppforlife, line 25: why is token needed here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+        run: |
+          set -e -x
+          cat ./tmp/checksums.txt
+          diff ./tmp/checksums.txt <(cat <<EOF
+          ${{steps.get-checksums-from-draft-release.outputs.result}}
+          EOF
+          )
+      - name: Run PackageCR Validation Tests
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
cppforlife: is token needed here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+            serviceAccountName: cluster-admin-sa
+            packageRef:
+              refName: secretgen-controller.carvel.dev
+              apiVersion: packaging.carvel.dev/v1alpha1
+              versionSelection:
+                constraints: "${constraintVersion:1}"
+          EOF
+
+          # download the generated package and metadata from github draft release
+          gh release download "${{ github.ref_name }}" -p "*.yml"
```
cppforlife: could we just use files that we generated in previous steps?

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+      - name: Create release draft and upload release yaml
+        uses: softprops/action-gh-release@17cd0d34deddf848fc0e7d9be5202c148c270a0a
+        with:
+          name: ${{ github.ref_name }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          body_path: ./tmp/checksums.txt
+          files: |
+            ./tmp/release.yml
+            ./tmp/package.yml
+            ./tmp/metadata.yml
```
cppforlife: might be worth calling this `package-metadata.yml` (metadata.yml looks a bit out of context without the prefix)

.github/workflows/release.yml
neil-hickey: yes, we can change that out. This was previously two different jobs, and so it needed a upload / download step to get the files again. Nice catch

.github/workflows/release.yml
neil-hickey: From: https://github.com/vmware-tanzu/carvel-setup-action#usage
```
carvel-setup-action uses the GitHub API to find information about latest releases. To avoid rate limits it is recommended you pass a token:
```

.github/workflows/release.yml
neil-hickey: It's named `metadata.yml` by the `build-release.sh` script. Might have to edit it there as part of this PR to get this change fully resolved.

.github/workflows/release.yml
cppforlife: i see, ok. we ll try to resolve this eventually via carvel-release-scripts.

.github/workflows/release.yml
neil-hickey: For consistency sake, I went through and renamed metadata.yml to package-metadata

I prefer the ability to work back from the release artifact by name to the file.

PR #94, opened 2022-07-25
Package Authoring GitHub Action SGC

sethiyash:
- Github action to package SGC using package authoring CLI


.github/workflows/package-release.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Setting up minikube
+        run: |
+          set -e -x
+          minikube start --driver=docker --wait=all
+          docker buildx create minikube --use --driver=kubernetes --bootstrap
+
+      - name: Fetch kctrl
+        uses: actions/checkout@v3
+        with:
+          ref: kctrl-pkg-init-release
```
sethiyash: Currently, Pointing to the branch where we have the changes. Will update it once the changes got merged.

package-build.yml
```diff
@@ -0,0 +1,29 @@
...
+          deploy:
+          - kapp: {}
+          template:
+          - ytt:
+              paths:
+              - config/package-bundle/config
+              - config/release
+          - kbld:
+              paths:
+              - config/release
```
rohitagg2020: Need to add '-' instead of `config/release`

package-build.yml
```diff
@@ -0,0 +1,29 @@
...
+      export:
+      - imgpkgBundle:
+          image: ysethiya/testsgc
```
rohitagg2020: Needs to be changed.

package-build.yml
```diff
@@ -0,0 +1,29 @@
+apiVersion: kctrl.carvel.dev/v1alpha1
+kind: PackageBuild
+metadata:
+  creationTimestamp: "2022-07-21T18:41:22Z"
```
rohitagg2020: I think we can remove creationTimestamp.

rohitagg2020:
Package-resources.yml file is missing. It should have values similar to those mentioned in package and package-metadata(inside config/package folder) otherwise these files when generated by kctrl will have different values.

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,64 @@
+#release.yml
```
praveenrewar: nit: Space after `#`

.github/workflows/package-release.yml
praveenrewar: I believe a lot of these steps will go away once we have released kctrl with package authoring because kctrl will get installed as part of the GH action (carvel-setup-action@v1)

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,64 @@
...
+          ./carvel-kapp-controller/cli/kctrl pkg release -y -v ${{ github.ref_name }} --debug
+      
+      - name: Create draft release with package and metadata
+        uses: softprops/action-gh-release@17cd0d34deddf848fc0e7d9be5202c148c270a0a
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          files: |
+            carvel-artifacts/packages/secretgen-controller.carvel.dev/metadata.yml
+            carvel-artifacts/packages/secretgen-controller.carvel.dev/package.yml
+          draft: true
```
praveenrewar: nit: add newline at the end.

package-build.yml
sethiyash: Removed.

package-build.yml
sethiyash: Done.

.github/workflows/package-release.yml
sethiyash: Yup that's correct.

.github/workflows/package-release.yml
sethiyash: Done

.github/workflows/package-release.yml
sethiyash: removed this doesn't make much sense.

.github/workflows/package-release.yml
praveenrewar: I think you added one extra line :)

package-resources.yml
```diff
@@ -0,0 +1,54 @@
...
+
+---
+apiVersion: data.packaging.carvel.dev/v1alpha1
+kind: PackageMetadata
+metadata:
+  creationTimestamp: null
+  name: secretgen-controller.carvel.dev
+spec:
+  displayName: secretgen-controller
+  longDescription: secretgen-controller.carvel.dev
```
rohitagg2020: These values are different from what they are releasing with today. It has to be same.

.github/workflows/package-release.yml
```diff
@@ -0,0 +1,64 @@
...
```
rohitagg2020: There are test running on generated PackageCR in `release.yml` which is missing.

package-resources.yml
sethiyash: Updated PackageMetada

.github/workflows/package-release.yml
sethiyash: Modified release.yml itself to generate Package and PackageMetada using kctrl

hack/build-and-publish-release.sh
```diff
@@ -14,19 +14,4 @@ yq eval '.metadata.annotations."secretgen-controller.carvel.dev/version" = env(v
 
 ytt -f config/package-bundle/config -f config/release -v dev.version="$version_without_v_prefix" | kbld --imgpkg-lock-output config/package-bundle/.imgpkg/images.yml -f- > ./tmp/release.yml
 
-imgpkg push -b ghcr.io/vmware-tanzu/carvel-secretgen-controller-package-bundle:"$version" -f config/package-bundle --lock-output ./tmp/bundle-image.yml
```
neil-hickey, line 17: 🎉 ! this is great

neil-hickey:
See some generated files based on this workflow here: https://github.com/sethiyash/carvel-secretgen-controller/releases/tag/v4.5.7

package-resources.yml
```diff
@@ -0,0 +1,58 @@
+apiVersion: data.packaging.carvel.dev/v1alpha1
+kind: Package
+metadata:
+  name: secretgen-controller.carvel.dev.0.0.0
+spec:
```
neil-hickey, line 5: - Is there a way to get `releaseNotes: https://github.com/vmware-tanzu/carvel-secretgen-controller/releases/tag/v0.10.3` back in here?
- Also missing the `licenses: - Apache 2.0` section

package-resources.yml
```diff
@@ -0,0 +1,58 @@
...
+  longDescription: "Provide CRDs to specify what secrets need to be on cluster (generated or not)."
+  shortDescription: "Secret generation and sharing"
+  providerName: carvel.dev
+  maintainers:
+    - name: Carvel team (see www.carvel.dev for team members)
+  categories:
+    - "package management"
+
+---
+apiVersion: packaging.carvel.dev/v1alpha1
```
neil-hickey, line 42: What is the PKGI used for when running the release steps of kctrl? We don't publish it right? Or does this also include a PKGI to our release assets?

.github/workflows/release.yml
```diff
@@ -19,13 +19,6 @@ jobs:
           go-version: 1.18
       - name: Install Carvel Tools
         uses: vmware-tanzu/carvel-setup-action@v1
-        with:
```
neil-hickey, line 22: I think we have so far been very explicit about what versions we are releasing with. 

Our testing workflow does not float [see here](https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/.github/workflows/test-gh.yml#L28-L35). Keeping these in sync very explicitly means we know that the tested dependencies matches the released dependencies.

package-build.yml
```diff
@@ -0,0 +1,29 @@
...
+  template:
+    spec:
+      app:
+        spec:
+          deploy:
+          - kapp: {}
+          template:
+          - ytt:
+              paths:
+              - config/package-bundle/config
```
neil-hickey, line 17: I think one interesting consequence of this is that if another team (perhaps someone that integrates and adds overlays to SGC such as TKG) previously relied on the path being `config` and they have stuff they can just add to that folder before template stage. This just changes that path to be more specific. I wonder what the consequences are.

package-build.yml
```diff
@@ -0,0 +1,29 @@
...
+              - config/release
+          - kbld:
+              paths:
+              - '-'
+              - config/release
+      export:
+      - imgpkgBundle:
+          image: ghcr.io/vmware-tanzu/carvel-secretgen-controller
```
neil-hickey: Is there a way for this value to come from the release values  schema?  Seems like now we have some duplication?

https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/develop/config/release/values-schema.yml

sethiyash: So actually Neil this is how Package authoring CLI works as of now it reads whatever there is in `package-build.yml` and this file gets automatically generated when we do `kctrl pkg init` and during `kctrl pkg release` whatever registry URL inputted by the user it get store here.

package-resources.yml
sethiyash: This file got automatically generated when the user do kctrl pkg init and I believe we tried to keep as minimal fields as possible in the Package CR and don't generate which are not compulsory.

@rohitagg2020 if we add it manually here does that help?

package-resources.yml
sethiyash: No we won't publish it and is not part of release artifact

package-resources.yml
neil-hickey: I think we want to include the license section at a minimum. Whether that's done manually of via some flag to kctrl pkg int doesn't matter too much (I guess automation is always better though)

package-resources.yml
neil-hickey: so is it just something for kctrl to use? Or is it needed? I haven't dived into the kctrl flow just yet, so my questions are a little naive :)

sethiyash: Yeah it is something needed for kctrl

.github/workflows/release.yml
sethiyash: Got it. Using specific versions only now.

package-build.yml
sethiyash: @rohitagg2020 your thoughts on this?

package-resources.yml
sethiyash: Okay sure. We have figured out one bug that even if putting license section in package-resource.yml it does not reflect in final package.yml generated by kctrl. We will be resolving this from kctrl side

package-build.yml
sethiyash: @neil-hickey can you elaborate more on the scenario for which TKG currently apply overlays if you are aware of their flow?

package-build.yml
```diff
@@ -0,0 +1,28 @@
...
+    spec:
+      app:
+        spec:
+          deploy:
+          - kapp: {}
+          template:
+          - ytt:
+              paths:
+              - config/package-bundle/config
+              - tmp/output-config.yml
```
rohitagg2020: `tmp` folder in the imgpkg bundle doesn't sound right. Can we rename it?

package-build.yml
```diff
@@ -0,0 +1,28 @@
...
+          - kbld:
+              paths:
+              - '-'
+      export:
+      - imgpkgBundle:
+          image: ghcr.io/vmware-tanzu/carvel-secretgen-controller
+          useKbldImagesLock: true
+        includePaths:
+        - config/package-bundle/config
+        - tmp/output-config.yml
```
rohitagg2020: `tmp` folder in the imgpkg bundle doesn't sound right. Can we rename it?

package-build.yml
```diff
@@ -0,0 +1,28 @@
...
+          deploy:
+          - kapp: {}
+          template:
+          - ytt:
+              paths:
+              - config/package-bundle/config
+              - tmp/output-config.yml
+          - kbld:
+              paths:
+              - '-'
```
rohitagg2020, line 21: Did we check how to handle generating two images while making a release?

sethiyash: Yeah as discussed now I have started passing images.yml in kbld step.

package-build.yml
sethiyash: yeah now putting it inside config folder only.

package-resources.yml
```diff
@@ -17,7 +17,7 @@ spec:
       - ytt:
           paths:
           - config/package-bundle/config
-          - tmp/output-config.yml
+          - config/relase-config.yml
```
rohitagg2020: `release` spelling is misspelled as `relase`

package-build.yml
```diff
@@ -15,14 +15,16 @@ spec:
           - ytt:
               paths:
               - config/package-bundle/config
-              - tmp/output-config.yml
+              - config/relase-config.yml
```
rohitagg2020: release spelling is misspelled as relase

.github/workflows/release.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       - name: Run Package build
         run: |
           constraintVersion="${{ github.ref_name }}"
-          ytt -f config/release > ./tmp/output-config.yml
+          ytt -f config/release > ./config/relase-config.yml
```
rohitagg2020: release spelling is misspelled as relase

sethiyash: Refactored.

package-build.yml
sethiyash: Refactored

package-resources.yml
sethiyash: Refactored

.github/workflows/release.yml
```diff
@@ -46,8 +58,8 @@ jobs:
           body_path: ./tmp/checksums.txt
           files: |
             ./tmp/release.yml
-            ./tmp/package.yml
-            ./tmp/package-metadata.yml
+            carvel-artifacts/packages/secretgen-controller.carvel.dev/package.yml
+            carvel-artifacts/packages/secretgen-controller.carvel.dev/metadata.yml
```
rohitagg2020: In today `secretgen-controller` releases, file is called package-metadata.yml

rohitagg2020: Also, if we see the checksum in already releases, it is something like this:
```
5c95408656d898f724df7a29ec4052762a94711a8677637625fbb770af2f3cf4  release.yml
aa1326743e61eb8e9e1122668a9434961a47c36982b65e4d33cc08eab0618dd5  package.yml
55338cd490a6932c857e2035ed552b0a21e07bd2f0a46d4f5d0d2419b2323333  package-metadata.yml
```

neil-hickey: I wouldn't consider the naming a blocker. Given that this is a release artifact I just wanted it be very explicit that it was a package metadata file and not some other kind of release metadata. For the purposes of functionality, it shouldn't matter too much. Anyone vendiring the file's will probably have to update, but that's ok. I don't think there are many people doing that.

package-build.yml
```diff
@@ -0,0 +1,30 @@
...
+              paths:
+              - config/package-bundle/config
+              - config/release-config.yml
+          - kbld:
+              paths:
+              - '-'
+              - .imgpkg/images.yml
+      export:
+      - imgpkgBundle:
+          image: ghcr.io/vmware-tanzu/carvel-secretgen-controller
```
neil-hickey: So I think this is where my final question is.

Today we publish two things:
1. An image for secretgen-controller
2. A bundle with all the  config to deploy secretgen-controller as a Package/PackageInstall

We therefore have two packages: https://github.com/orgs/vmware-tanzu/packages?repo_name=carvel-secretgen-controller 

In the previous world, we had the value [here](https://github.com/vmware-tanzu/carvel-secretgen-controller/blob/2601d33519a61278c258a3664f4f39f49bcfe750/config/release/values-schema.yml#L8) for the image location, and I believe as part of the release this was published. Is this behaviour maintained? I am just seeing the `bundle` package being published here, I am just unsure of whether the image is still being published or not now? I think some users rely on just the image without the bundle aspect. So I'd like to maintain those separately. Right now it looks like we are changing `ghcr.io/vmware-tanzu/carvel-secretgen-controller` to be a bundle whereas right now it's an image.

Hopefully this is clear.

sethiyash: As discussed. It still publishes both things. Also, I updated imgpkg bundle registry URL to - `ghcr.io/vmware-tanzu/carvel-secretgen-controller-package-bundle` as discussed.

.github/workflows/release.yml
sethiyash: Changed the file name to `package-metadata.yml` as it was earlier.

neil-hickey (approved):
LGTM! Thanks for making the changes 👍

cppforlife (approved):

PR #99, opened 2022-08-17
fixes several oversights and maybe a real vuln from trivvy

joe-kimmel-vmw:
- Trivy was still running tests against golang 1.17 so this upgrades trivy and all the other places i could find to golang 1.18.  
- see e.g. all the vulns from this run: https://github.com/vmware-tanzu/carvel-secretgen-controller/security/code-scanning
- upgrades prometheus client library just in case that mattered
- then there was some github-runner script error which idk why it appeared magically just now but maybe i made a change that fixed it, or at least didn't break anything any further?  https://github.com/vmware-tanzu/carvel-secretgen-controller/runs/7870769199?check_suite_focus=true#step:8:24
- then there was that other thing so i upgraded that too: https://github.com/vmware-tanzu/carvel-secretgen-controller/runs/7870999891?check_suite_focus=true#step:7:11

### DIFF
--- a/.github/workflows/trivy-scan.yml
+++ b/.github/workflows/trivy-scan.yml
@@ -92,7 +92,7 @@ jobs:
           trivy image --format template --output trivy-results-image.sarif "docker.io/carvel/secretgen-controller:${{ github.sha }}"
           trivy image --format json --output trivy-results-image.json "docker.io/carvel/secretgen-controller:${{ github.sha }}"
       - name: Upload Trivy scan results to GitHub Security tab
-        uses: github/codeql-action/upload-sarif@v1
+        uses: github/codeql-action/upload-sarif@v2
         with:
           sarif_file: '.'
       - name: Check for new Vulnerabilities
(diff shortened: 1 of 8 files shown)


.github/workflows/trivy-scan.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - name: Set up Go 1.x
         uses: actions/setup-go@v1
         with:
-          go-version: "1.17"
+          go-version: "1.18"
```
joaopapereira: If we are using a specific version of go in kind-action we should use it here as well. If not we might be running scans on different versions of the stdlib

.github/workflows/trivy-scan.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - name: Set up Go 1.x
         uses: actions/setup-go@v1
```
joaopapereira:
```suggestion
        uses: actions/setup-go@v3
```
Let us maybe use this opportunity to align the version of the action with

joaopapereira (commented):
Left some comments

cppforlife:
lgtm

PR #104, opened 2022-08-20
Say "secretgen-controller" in install instructions

mamachanko:
because I think that's what's meant.


neil-hickey:
LGTM :) Thanks @mamachanko !

PR #114, opened 2022-11-08
Remove TCE steps from release checklist

neil-hickey:


joaopapereira (approved):
LGTM

neil-hickey:
Turns out for now we still need to do this. Will re-open when we know the TCE stuff is gone forever.

PR #133, opened 2022-11-10
Port base OS image change from photon to scratch

neil-hickey:
This will reduce our CVE burden greatly by not including a full OS where it's not needed.


neil-hickey:
Mostly copied from https://github.com/vmware-tanzu/carvel-secretgen-controller/pull/76/

joaopapereira (commented):
This sounds like a great idea :D LGTM

PR #148, opened 2022-11-16
resolve #110: Add extra requirements on the Password CRD

### LINKED ISSUE #110: Add extra requirements on the Password CRD.
The `secretgen.k14s.io/v1alpha1/Password` allows to specify the length of the generated password only. Unfortunately, external resources may ask to have other requirements such as: `Your password must contain characters from three of the following categories – English uppercase letters, English lowercase letters, numbers (0-9), and non-alphanumeric characters (!, $, #, %, etc.).` or `Your password must be at least 8 characters and at most 128 characters`

Add extra properties to manage these requirements. The password crd can look like
```yaml
apiVersion: secretgen.k14s.io/v1alpha2
kind: Password
metadata:
  name: my-constrained-long-user-password
spec:
  length:
      min: 8
      max: 128
  categories:
     - name: uppercase        #pre-defined category
        min: 2
     - name: lowercase        #pre-defined category + min:1 default values                                              
     - name: my-custom-set
        min: 3
        available: ['!', '$', '#', '%']
```
---
Vote on this request

bmoussaud:
Implement issue #110 to be able to provide more inputs on the specification of the generated password.

### DIFF
--- a/pkg/apis/secretgen/v1alpha1/types_password.go
+++ b/pkg/apis/secretgen/v1alpha1/types_password.go
@@ -47,7 +47,28 @@ type PasswordList struct {
 
 type PasswordSpec struct {
 	// +optional
+	// +kubebuilder:default:=40
 	Length int `json:"length"`
+
+	// +optional
+	// +kubebuilder:default:=0
+	Digits int `json:"digits"`
+
+	// +optional
+	// +kubebuilder:default:=0
+	Symbols int `json:"symbols"`
+
+	// +optional
+	// +kubebuilder:default:=0
+	UppercaseLetters int `json:"uppercaseLetters"`
+
+	// +optional
+	// +kubebuilder:default:=0
+	LowercaseLetters int `json:"lowercaseLetters"`
+
+	// +optional
+	// +kubebuilder:default:="!@#$%&*;.:"
+	SymbolCharSet string `json:"symbolCharSet"`
 
 	// +optional
 	SecretTemplate *SecretTemplate `json:"secretTemplate,omitempty"`
--- a/pkg/generator/password_reconciler.go
+++ b/pkg/generator/password_reconciler.go
@@ -5,9 +5,13 @@ package generator
 
 import (
 	"context"
+	"crypto/rand"
 	"fmt"
+	"math/big"
+	"strings"
 
-	cfgtypes "github.com/cloudfoundry/config-server/types"
+	mathrand "math/rand"
+
 	"github.com/go-logr/logr"
 	sgv1alpha1 "github.com/vmware-tanzu/carvel-secretgen-controller/pkg/apis/secretgen/v1alpha1"
 	sgclient "github.com/vmware-tanzu/carvel-secretgen-controller/pkg/client/clientset/versioned"
@@ -29,6 +33,12 @@ type PasswordReconciler struct {
 }
 
 var _ reconcile.Reconciler = &PasswordReconciler{}
+
+var (
+	lowerCharSet = "abcdefghijklmnopqrstuvwxyz"
+	upperCharSet = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	digitSet     = "0123456789"
+)
 
 func NewPasswordReconciler(sgClient sgclient.Interface,
 	coreClient kubernetes.Interface, log logr.Logger) *PasswordReconciler {
@@ -115,37 +125,75 @@ func (r *PasswordReconciler) createSecret(ctx context.Context, password *sgv1alp
 	return reconcile.Result{}, nil
 }
 
-type passwordParams struct {
-	Length int `yaml:"length"`
+func (r *PasswordReconciler) generate(password *sgv1alpha1.Password) (string, error) {
+	passwordVal := r.GeneratePassword(&password.Spec)
+	return passwordVal, nil
 }
 
-func (r *PasswordReconciler) generate(password *sgv1alpha1.Password) (string, error) {
-	gen := cfgtypes.NewPasswordGenerator()
+// https://golangbyexample.com/generate-random-password-golang/
+func (r *PasswordReconciler) GeneratePassword(spec *sgv1alpha1.PasswordSpec) string {
 
-	genParams := passwordParams{Length: password.Spec.Length}
-	if genParams.Length == 0 {
-		genParams.Length = 40
+	var password strings.Builder
+
+	//Set symbol character
+	for i := 0; i < spec.Symbols; i++ {
+		value, _ := randomElement(spec.SymbolCharSet)
+		password.WriteString(value)
 	}
 
-	passwordVal, err := gen.Generate(genParams)
+	//Set digit
+	for i := 0; i < spec.Digits; i++ {
+		value, _ := randomElement(digitSet)
+		password.WriteString(value)
+	}
+
+	//Set uppercase
+	for i := 0; i < spec.UppercaseLetters; i++ {
+		value, _ := randomElement(upperCharSet)
+		password.WriteString(value)
+	}
+
+	//Set lowercase
+	for i := 0; i < spec.LowercaseLetters; i++ {
+		value, _ := randomElement(lowerCharSet)
+		password.WriteString(value)
+	}
+
+	var allCharSet = lowerCharSet + upperCharSet + digitSet
+
+	remainingLength := spec.Length - spec.Symbols - spec.Digits - spec.UppercaseLetters - spec.LowercaseLetters
+	for i := 0; i < remainingLength; i++ {
+		value, _ := randomElement(allCharSet)
+		password.WriteString(value)
+	}
+
+	inRune := []rune(password.String())
+	mathrand.Shuffle(len(inRune), func(i, j int) {
+		inRune[i], inRune[j] = inRune[j], inRune[i]
+	})
+
+	return string(inRune)
+}
+
+func randomElement(s string) (string, error) {
+	n, err := rand.Int(rand.Reader, big.NewInt(int64(len(s))))
 	if err != nil {
 		return "", err
 	}
-
-	return passwordVal.(string), nil
+	return string(s[n.Int64()]), nil
 }
 
 func (r *PasswordReconciler) updateStatus(ctx context.Context, password *sgv1alpha1.Password) error {
 	existingPassword, err := r.sgClient.SecretgenV1alpha1().Passwords(password.Namespace).Get(ctx, password.Name, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("Fetching password: %s", err)
+		return fmt.Errorf("fetching password: %s", err)
 	}
 
 	existingPassword.Status = password.Status
 
 	_, err = r.sgClient.SecretgenV1alpha1().Passwords(existingPassword.Namespace).UpdateStatus(ctx, existingPassword, metav1.UpdateOptions{})
 	if err != nil {
-		return fmt.Errorf("Updating password status: %s", err)
+		return fmt.Errorf("updating password status: %s", err)
 	}
 
 	return nil
--- a/pkg/generator/password_reconciler.go
+++ b/pkg/generator/password_reconciler.go
@@ -32,12 +32,6 @@ type PasswordReconciler struct {
 
 var _ reconcile.Reconciler = &PasswordReconciler{}
 
-var (
-	lowerCharSet = "abcdefghijklmnopqrstuvwxyz"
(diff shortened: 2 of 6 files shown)


pkg/generator/password_reconciler.go
```diff
@@ -115,37 +125,75 @@ func (r *PasswordReconciler) createSecret(ctx context.Context, password *sgv1alp
...
+	var allCharSet = lowerCharSet + upperCharSet + digitSet
+
+	remainingLength := spec.Length - spec.Symbols - spec.Digits - spec.UppercaseLetters - spec.LowercaseLetters
+	for i := 0; i < remainingLength; i++ {
+		value, _ := randomElement(allCharSet)
+		password.WriteString(value)
+	}
+
+	inRune := []rune(password.String())
+	mathrand.Shuffle(len(inRune), func(i, j int) {
```
cppforlife: math/rand unfortunately is "unsuitable for security-sensitive work" per math/rand's package documentation. even though we are using crypto/rand to select individual elements, i think shuffling needs to be "secure" as well.

bmoussaud:
@cppforlife can you merge it?

cppforlife:
i ve recently nudged folks to give a second set of eyes; seems like folks havent gotten any additional feedback. let me take a second look (first look i did, it looked fine to me) and we'll get it merged.

pkg/apis/secretgen/v1alpha1/types_password.go
```diff
@@ -47,8 +47,29 @@ type PasswordList struct {
...
+	// +kubebuilder:default:=0
+	UppercaseLetters int `json:"uppercaseLetters"`
+
+	// +optional
+	// +kubebuilder:default:=0
+	LowercaseLetters int `json:"lowercaseLetters"`
+
+	// +optional
+	// +kubebuilder:default:="!@#$%&*;.:"
+	SymbolCharSet string `json:"symbolCharSet"`
```
cppforlife, line 71: nit: it seems that it's more common to not have `charset` be "one word", meaning we should rename this key to `symbolCharset`. wdyt?

pkg/apis/secretgen/v1alpha1/types_password.go
```diff
@@ -47,8 +47,29 @@ type PasswordList struct {
 
 type PasswordSpec struct {
 	// +optional
+	// +kubebuilder:default:=40
```
cppforlife, line 50: i dont think we use kubebuilder in the project, so to avoid confusion that we do i would recommend removing these comments. if we want to indicate defaults, perhaps do it in a description (comment without qualifier)?

pkg/generator/password_reconciler.go
```diff
@@ -115,37 +123,96 @@ func (r *PasswordReconciler) createSecret(ctx context.Context, password *sgv1alp
...
-type passwordParams struct {
-	Length int `yaml:"length"`
+func (r *PasswordReconciler) generate(password *sgv1alpha1.Password) (string, error) {
+	passwordVal := localGeneratePassword(&password.Spec)
+	return passwordVal, nil
 }
 
-func (r *PasswordReconciler) generate(password *sgv1alpha1.Password) (string, error) {
-	gen := cfgtypes.NewPasswordGenerator()
+func localGeneratePassword(spec *sgv1alpha1.PasswordSpec) string {
```
cppforlife: nit: inline localGeneratePassword into  PasswordReconciler.generate -- it's not clear to me why it's needed to be separated.

pkg/generator/password_reconciler.go
```diff
@@ -115,37 +123,96 @@ func (r *PasswordReconciler) createSecret(ctx context.Context, password *sgv1alp
...
+
+	var password strings.Builder
+
+	//Set symbol character
+	for i := 0; i < spec.Symbols; i++ {
+		value, _ := randomElement(spec.SymbolCharSet)
```
cppforlife: seems like we we are ignoring errors when generating elements, i think they should be propagated since failure to generate char shouldnt be ignored.

pkg/generator/password_reconciler.go
```diff
@@ -30,6 +32,12 @@ type PasswordReconciler struct {
 
 var _ reconcile.Reconciler = &PasswordReconciler{}
 
+var (
```
cppforlife: can be const. also allCharSet can be next to these as well.

cppforlife: i would also recommend moving const declaration into generate function. e.g.

```go
func (r *PasswordReconciler) generate(password *sgv1alpha1.Password) (string, error) {
   const (
      lowerCharset = "..."
   )
   ...other code...
}
```

this way it's localized to where content is actually needed.

joe-kimmel-vmw (approved):
LGTM

pkg/generator/password_reconciler.go
```diff
@@ -115,37 +123,96 @@ func (r *PasswordReconciler) createSecret(ctx context.Context, password *sgv1alp
...
-	if genParams.Length == 0 {
-		genParams.Length = 40
+	//Set lowercase
+	for i := 0; i < spec.LowercaseLetters; i++ {
+		value, _ := randomElement(lowerCharSet)
+		password.WriteString(value)
 	}
 
-	passwordVal, err := gen.Generate(genParams)
+	var allCharSet = lowerCharSet + upperCharSet + digitSet
```
neil-hickey, line 159: Is this missing the symbolcharset?

pkg/generator/password_reconciler.go
```diff
@@ -115,37 +123,96 @@ func (r *PasswordReconciler) createSecret(ctx context.Context, password *sgv1alp
...
-		return fmt.Errorf("Fetching password: %s", err)
+		return fmt.Errorf("fetching password: %s", err)
 	}
 
 	existingPassword.Status = password.Status
 
 	_, err = r.sgClient.SecretgenV1alpha1().Passwords(existingPassword.Namespace).UpdateStatus(ctx, existingPassword, metav1.UpdateOptions{})
 	if err != nil {
-		return fmt.Errorf("Updating password status: %s", err)
+		return fmt.Errorf("updating password status: %s", err)
```
neil-hickey, line 224: Question for other maintainers, i.e @joe-kimmel-vmw / @cppforlife seems like we have a mix of upper and lower case error formatting going on. I think go conventionally pushes for the lower case, but I am see a LOT of upper cased error messaging in this repo. Any thoughts on  why we went with the uppercasing generally?

pkg/generator/password_reconciler.go
neil-hickey, line 165: Is this missing the symbolcharset? 

```suggestion
	var allCharSet = lowerCharSet + upperCharSet + digitSet + spec.SymbolCharSet
```

test/e2e/password_test.go
```diff
@@ -138,3 +138,217 @@ spec:
...
+spec:
+  length: 24
+  secretTemplate:
+    type: Opaque
+    stringData:     
+      val: $(value)
+      valx2: $(value)$(value)
+`
+
+	name := "test-lenght-password-template"
```
neil-hickey:
```suggestion
	name := "test-length-password-template"
```

test/e2e/password_test.go
```diff
@@ -138,3 +138,217 @@ spec:
...
+		if err != nil {
+			t.Fatalf("Failed to unmarshal: %s", err)
+		}
+
+		if secret.Type != "Opaque" {
+			t.Fatalf("Wrong type")
+		}
+
+		if len(secret.Data["val"]) != 24 {
+			t.Fatalf("Expect password lenght is 24, found %d", len(secret.Data["val"]))
```
neil-hickey:
```suggestion
			t.Fatalf("Expect password length is 24, found %d", len(secret.Data["val"]))
```

test/e2e/password_test.go
```diff
@@ -138,3 +138,217 @@ spec:
...
+		}
+		if len(secret.Data["valx2"]) != 2*24 {
+			t.Fatalf("Expect valx2 password lenght is 48, found %d", len(secret.Data["valx2"]))
```
neil-hickey:
```suggestion
			t.Fatalf("Expect valx2 password length is 48, found %d", len(secret.Data["valx2"]))
```

test/e2e/password_test.go
```diff
@@ -138,3 +138,217 @@ spec:
...
+		if err != nil {
+			t.Fatalf("Failed to unmarshal: %s", err)
+		}
+
+		if secret.Type != "Opaque" {
+			t.Fatalf("Wrong type")
+		}
+
+		if len(secret.Data["val"]) != 27 {
+			t.Fatalf("Expect password lenght is 27, found %d (value %s) ", len(secret.Data["val"]), secret.Data["val"])
```
neil-hickey:
```suggestion
			t.Fatalf("Expect password length is 27, found %d (value %s) ", len(secret.Data["val"]), secret.Data["val"])
```

test/e2e/password_test.go
```diff
@@ -138,3 +138,217 @@ spec:
...
+		if err != nil {
+			t.Fatalf("Failed to unmarshal: %s", err)
+		}
+
+		if secret.Type != "Opaque" {
+			t.Fatalf("Wrong type")
+		}
+
+		if len(secret.Data["val"]) != 3 {
+			t.Fatalf("Expect password lenght is 3, found %d", len(secret.Data["val"]))
```
neil-hickey:
```suggestion
			t.Fatalf("Expect password length is 3, found %d", len(secret.Data["val"]))
```

test/e2e/password_test.go
```diff
@@ -138,3 +138,217 @@ spec:
...
+		err := yaml.Unmarshal([]byte(out), &secret)
+		if err != nil {
+			t.Fatalf("Failed to unmarshal: %s", err)
+		}
+
+		if secret.Type != "Opaque" {
+			t.Fatalf("Wrong type")
+		}
+
+		if len(secret.Data["val"]) != 3 {
```
neil-hickey: We use the `"github.com/stretchr/testify/assert"` package to test generally in SGC. (I appreciate that the code has a mixture, but are moving as a project to add new code with this package used instead of t.Fatal etc)

This would actually simply alot of this testing to something like:

```suggestion
		assert.Len(t, secret.Data["val"], 3, "Expect password length is 3, found %d", len(secret.Data["val"])")
```

pkg/generator/password_reconciler.go
bmoussaud: hey, @neil-hickey: I didn't add spec.SymbolCharSet because I don't want to add extra SymbolCharset to fill in the generated password. Systems don't allow too many symbols. So if the user specifies x symbols, the user gets x symbols. not more.

neil-hickey: ok cool, seems reasonable to me. Just wanted to double check the reasoning 👍

test/e2e/password_test.go
```diff
@@ -138,3 +138,217 @@ spec:
...
+		err := yaml.Unmarshal([]byte(out), &secret)
+		if err != nil {
+			t.Fatalf("Failed to unmarshal: %s", err)
+		}
+
+		if secret.Type != "Opaque" {
+			t.Fatalf("Wrong type")
+		}
+
+		assert.Len(t, secret.Data["val"], 3, "Expect password length is 3, found %d", len(secret.Data["val"])")
```
neil-hickey:
```suggestion
		assert.Len(t, secret.Data["val"], 3, "Expect password length is 3, found %d", len(secret.Data["val"]))
```

I think my suggestion was bad, left in an extra `"`

pkg/generator/password_reconciler.go
cppforlife: carvel projects always used capital letter for starting errors.

cppforlife: > Any thoughts on why we went with the uppercasing generally?

i found it easier to read when k14s was started and also it was easier to split errors to be more human friendly.

cppforlife (commented):
(seems like i had pending comments on this PR, probably waiting for others to reply first. submitted now.)

pkg/generator/password_reconciler.go
joe-kimmel-vmw: +1 - it's initial caps because that's what DK prefers.  I think i tried to put that in our mostly-useless style guide at some point.

pkg/generator/password_reconciler.go
bmoussaud: @cppforlife is it mandatory to apply this pattern before merging the pull request?

cppforlife (approved):
note to self: post-merge add err check for rand.Int calls

PR #150, opened 2022-11-18
Introduce inclusive naming check and update existing language

aaronshurley:
We are following the Inclusive Naming Initiative's guidance as that is what the CNCF supports.


neil-hickey (approved):
LGTM

PR #158, opened 2022-11-28
Bump golang to 1.19.3

neil-hickey:


neil-hickey:
I'm not sure why the checks aren't running... I ran the `test-gh` and `linter` locally. Both work fine...

PR #201, opened 2023-01-17
Removing build config from the SGC Package

sethiyash:


neil-hickey (approved):
LGTM, I see you validated on your fork that the release values are still used when you remove this, so that's great. Thanks!

PR #224, opened 2023-01-24
check for rand.Int error added via latest PR

cppforlife:


pkg/generator/password_reconciler.go
```diff
@@ -166,30 +166,33 @@ func (r *PasswordReconciler) generate(password *sgv1alpha1.Password) (string, er
...
-		return "", err
+		return "", fmt.Errorf("Getting rand int: %s", err)
 	}
 	return string(s[n.Int64()]), nil
 }
 
-func localCryptoShuffle(n int, swap func(i, j int)) {
+func (*PasswordReconciler) localCryptoShuffle(n int, swap func(i, j int)) error {
 	if n < 0 {
 		panic("invalid argument to Shuffle")
```
joaopapereira, line 197: This should not panic the generator unless it is an internal inconsistency error. This looks like invalid input from the spec itself, since we now have an output error maybe we can return it instead.

joaopapereira (commented):
The PR looks good I just have a minor comment that maybe we can improve.

neil-hickey:
hey @joaopapereira / @cppforlife, just checking in to see if we can push this PR through? Any blockers or discussions needed?

neil-hickey:
Hey @cppforlife / @joaopapereira . This one has been sitting for a long time, shall we close it out? Or can we get this over the line?

PR #262, opened 2023-03-03
Bump go version to 1.20.1

rohitagg2020:
- Bump go version to 1.20.1
- Bump golangci-lint to 1.51.2 as go 1.20 support was added from 1.51.0 onwards


neil-hickey:
I believe we are already past this version, so closing this one out

PR #335, opened 2023-05-09
Bumping go version to v1.20.3 and making golint happy

sethiyash:
- Bumping go version to v1.20.3 and making golint happy


praveenrewar (approved):
LGTM!

PR #337, opened 2023-05-11
feat: match ns annotations

### LINKED ISSUE #348: Feature: match allow on namespace annotations
We have multi-tenant clusters where isolation is done by rancher. The mechanism is simple, anyone is scoped in a rancher project, anyone can create new namespace and scope it in it's project scope by adding an annotation that is matching the project id. To summarize this annotation is the guarantee that this namespace will not be accessed by another tenant.
In our dev cluster, we have unpredictable new namespace that will be created for review environments and we want be able to use SecretImport in them, but the constraint in SecretExport based on the destination namespace is not sufficient to address our case.

We would like be able to specify `ToNamespaceAnnotations` as following, instead of `ToNamespace` on `SecretExport`
```yaml
kind: SecretExport
metadata:
  name: user-password
  namespace: user1
spec:
  toNamespaceAnnotations:
    projectId: my-project-id
```

This could also be
```yaml
kind: SecretExport
metadata:
  name: user-password
  namespace: user1
spec:
  toNamespaceAnnotationKey: projectId
  toNamespaceAnnotationValue: my-project-id
```

I can try to make a PR if this feature is welcome :-)

---
Vote on this request

devthejo:
fix #348

### DIFF
--- a/config/package-bundle/config/crds.yml
+++ b/config/package-bundle/config/crds.yml
@@ -40,6 +40,12 @@ spec:
                 items:
                   type: string
                 type: array
+              toNamespaceAnnotation:
+                type: object
+                additionalProperties: true
+              toNamespaceAnnotations:
+                type: object
+                additionalProperties: true
             type: object
           status:
             properties:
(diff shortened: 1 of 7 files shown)


devthejo:
closed in favor of #376

PR #375, opened 2023-05-30
Fix build arg in Dockerfile

### LINKED ISSUE #359: arm64 support for secretgen-controller
Hello,

I'm an engineer from the RabbitMQ team. Our commercial operator is using the carvel toolchain and the tanzu-cluster-essential components (kapp and secret-gen controller) as prerequisite.

We are trying to produce a commercial arm64 image to solve a limitation on tanzu-net not supporting multi-arch image.

I can see that the tanzu-cluster-essential at the moment don't support multi-arch images commercially (probably for the same limitation on tanzunet registry).

What about the open source versions? I can see that I can install kapp on arm64
 kapp deploy -a kc -f https://github.com/vmware-tanzu/carvel-kapp-controller/releases/latest/download/release.yml

but not secretgen controller:
kapp deploy -a sg -f https://github.com/carvel-dev/secretgen-controller/releases/latest/download/release.yml

where it seems is using the amd64 image.

Is there a way we can install secret-gen controller on arm64?

Thank you!

praveenrewar:
Fixes: #359

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /workspace
 
 COPY . .
 # helpful ldflags reference: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications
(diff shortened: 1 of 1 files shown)


neil-hickey (approved):
LGTM, ty!

PR #405, opened 2023-06-16
Get the latest tag during release

sethiyash:
- Made changes to version-util.sh script to get the latest tag always

### DIFF
--- a/hack/version-util.sh
+++ b/hack/version-util.sh
@@ -6,7 +6,7 @@
 
 # git tag version extraction adapted from https://github.com/carvel-dev/imgpkg/blob/develop/hack/build-binaries.sh
 function get_latest_git_tag {
-  git describe --tags | grep -Eo 'v[0-9]+\.[0-9]+\.[0-9]+(-alpha\.[0-9]+)?'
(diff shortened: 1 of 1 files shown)


hack/version-util.sh
```diff
@@ -6,13 +6,13 @@
...
 }
 
 function get_sgctrl_ver {
   echo "${1:-`get_latest_git_tag`}"
 }
 
 function get_sgctrl_ver_without_v {
-  git describe --tags | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+(-alpha\.[0-9]+)?'
+  git tag -l | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+(-alpha\.[0-9]+)?' | sort -V | tail -1
 }
```
rohitagg2020: nit: It seems like there is an extra line in the end. Can we remove extra line ? 
Otherwise LGTM

sethiyash: I think It's due to line missing I think golint expects to have new line at the end of the file

praveenrewar (approved):

hack/version-util.sh
sethiyash: But it was not earlier as well but I can add that

PR #406, opened 2023-06-18
Logging errors at get checksums stage in GH release action

sethiyash:
- Started logging errors at the `get uploaded release YAML checksums` stage in the GH release action to debug the failures while releasing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,9 +83,13 @@ jobs:
             const url = require('url');
             const https = require('https');
             checksums = {}
+            var releaseMatched = false;
+            var assetsFound = false;
             for (const r of releases["data"]) {
               if (r.draft && `refs/tags/${r.tag_name}` == "${{ github.ref }}") {
+                releaseMatched = true;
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -83,9 +83,13 @@ jobs:
             const url = require('url');
             const https = require('https');
             checksums = {}
+            var noReleaseMatched = true;
+            var noAssetsFound = true;
```
neil-hickey: So this comment *can* be ignored. But my own personal preference is to have variables be worded so that there isn't double negation --> i.e noAssetsFound = false is a little confusing to me. Whereas assetsFound = true / false is simple to understand.

sethiyash: Yeah! Definitely, that will be simple to understand. I will update it.

sethiyash: Adopted.

sethiyash:
@neil-hickey are we good to merge this now?

neil-hickey (approved):
LGTM, sorry I completely lost track of this Pr

PR #446, opened 2023-07-31
Decode input secrets used in template stringData

### LINKED ISSUE #445: Secrets used as input resources are double base64 encoded when used in stringData
**What steps did you take:**

1. Author a new SecretTemplate that uses stringData to build a multiline secret
2. Specify inputResources that are of kind Secret
3. Reference the inputResources in the stringData field
4. Apply the SecretTemplate

Example:
```
apiVersion: secretgen.carvel.dev/v1alpha1
kind: SecretTemplate
metadata:
  name: my-templated-secret
spec:
  inputResources:
  - name: my-input-secret
    ref:
      apiVersion: v1
      kind: Secret
      name: a-password-secret
  template:
    stringData:
      config: |
        #This is an example
        someKey = $(.my-input-secret.data.password)
```

**What happened:**
The values from the password secret were retrieved (base64 encoded) and added to the secret, the resulting string in the stringData field was then base64 encoded again. Meaning when my application decodes the templated secret, it doesn't contain the password, but rather the base64 contents of the original secrets. This makes the templated secret unusable.

**What did you expect:**
I expected the values to be base64 decoded so that my application could use them.

**Environment:**

- secretgen-controller version: v0.14.8
- Kubernetes version: v1.24.15

---
Vote on this request

We are also happy to receive and review Pull Requests if you want to help working on this issue.

carpenterm:
If an input resource is of kind Secret, and that is used in an expression in the stringData field of a SecretTemplate, then base64 decode it when evaluating the stringData field.

Fixes #445: Secrets used as input resources are double base64 encoded when used in stringData

### DIFF
--- a/pkg/generator/secret_template_reconciler.go
+++ b/pkg/generator/secret_template_reconciler.go
@@ -6,6 +6,7 @@ package generator
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -283,7 +284,12 @@ func evaluateTemplate(template *sg2v1alpha1.JSONPathTemplate, values map[string]
 	}
 
 	// Template Secret StringData
-	stringData, err := evaluate(template.StringData, values)
+	decodedValues, err := decodeSecrets(values)
+	if err != nil {
+		return corev1.Secret{}, fmt.Errorf("decoding secrets: %w", err)
+	}
+
+	stringData, err := evaluate(template.StringData, decodedValues)
 	if err != nil {
 		return corev1.Secret{}, fmt.Errorf("templating stringData: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/generator/secret_template_test.go
```diff
@@ -89,6 +89,62 @@ func Test_SecretTemplate(t *testing.T) {
...
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            "secretTemplate",
+					Namespace:       "test",
+					ResourceVersion: "1",
+					OwnerReferences: []metav1.OwnerReference{
+						secretTemplateOwnerRef("secretTemplate"),
+					},
+				},
+				Data: map[string][]byte{
```
neil-hickey, line 139: 139-142 should be covered by the previous unit-test.

pkg/generator/secret_template_test.go
```diff
@@ -89,6 +89,62 @@ func Test_SecretTemplate(t *testing.T) {
...
+					InputResources: []sg2v1alpha1.InputResource{{
+						Name: "creds",
+						Ref: sg2v1alpha1.InputResourceRef{
+							APIVersion: "v1",
+							Kind:       "Secret",
+							Name:       "existingSecret",
+						},
+					}},
+					JSONPathTemplate: &sg2v1alpha1.JSONPathTemplate{
+						Data: map[string]string{
```
neil-hickey, line 109: 109-112 should be covered by the previous unit-test. This test is only for StringData

pkg/generator/secret_template_reconciler.go
```diff
@@ -283,7 +284,12 @@ func evaluateTemplate(template *sg2v1alpha1.JSONPathTemplate, values map[string]
 	}
 
 	// Template Secret StringData
-	stringData, err := evaluate(template.StringData, values)
+	decodedValues, err := decodeSecrets(values)
```
neil-hickey, line 287: This is a pre-processing step which ends up having to iterate all the values even if you never specify a SecretData field. We also don't have any test for the "else" case, i.e.

```
if obj.GetKind() == "Secret" { } else { // no tests }
```



Is there a way to fold this into our existing `evaluate` function or a new function `evaluateStringData` which could delegate to the existing evaluateBytes or similar for most of the work.

devthejo:
I need this too, can I do something to help with this ?
@carpenterm are you still on the case ?

carpenterm:
Sorry but I am not working on this anymore, by the time the review came in I'd found another way using Kyverno.

PR #461, opened 2023-08-18
add cii badge to readme.md

microwavables:


praveenrewar (approved):

praveenrewar:
Closing and re-opening the PR as the tests seems to be hung up.

microwavables:
@praveenrewar checks still aren't triggering

praveenrewar:
Thanks to [this](https://github.com/orgs/community/discussions/26698#discussioncomment-3252954) comment, I realised that we had branch protection rules that require these 2 checks to pass before a PR can be merged, and we also have [some paths](https://github.com/carvel-dev/secretgen-controller/blob/55445d7aa286b254c45d05f9e56975e41aaaf8fd/.github/workflows/test-gh.yml#L8) added to path-ignore which would prevent these checks from being run for updated to markdown files which was conflicting. For now I have removed those branch protection rules (I think those were experimental as we don't have them in other tools) and the PR now says "All checks have passed" 🚀

PR #577, opened 2025-02-14
Fix CVEs: Bump go version to 1.23.6

devacts:


100mik (approved):
Good to go once the pipelines go green!

PR #709, opened 2025-11-25
Skip merging docker config with no auth data

devacts:
Fixes issue #710 

## PR Description:
When the registry does not require credentials, the source dockerconfigjson contains only the registry entry + certs, but no auth fields.

secretgen-controller merges this and produces e.g.

```
{
  "auths": {
    "harbor.internal": {
      "username": "",
      "password": "",
      "auth": ""
    }
  }
}
```

This is incorrect because the auths field is not empty, but the username, password, and auth fields are empty.
Kapp-controller then uses this placeholder secret for package/image fetching, registry rejects it with:
`Unsupported Auth config`

This PR updates the dockerconfigjson merge logic to skip registry entries that have no credentials, preventing sgc from generating invalid empty auth blocks that cause `Unsupported Auth config` errors with registry

### DIFF
--- a/pkg/sharing/dockerconfigjson_test.go
+++ b/pkg/sharing/dockerconfigjson_test.go
@@ -63,4 +63,28 @@ func Test_NewCombinedDockerConfigJSON(t *testing.T) {
 		assert.Equal(t, 1, len(result))
 		assert.Equal(t, expected, result[corev1.DockerConfigJsonKey])
 	})
+
+	t.Run("skips entries with no auth data", func(t *testing.T) {
+		// Secret with 1 valid and 1 empty auths
+		secrets := []*corev1.Secret{
+			&corev1.Secret{
+				Data: map[string][]byte{
+					corev1.DockerConfigJsonKey: []byte(`{"auths":{"valid-server1":{"username":"user1","password":"pass1","auth":"auth1"},"empty-server1":{"username":"","password":"","auth":""}}}`),
+				},
+			},
+			// Another Secret with 1 valid and 1 empty auths
+			&corev1.Secret{
+				Data: map[string][]byte{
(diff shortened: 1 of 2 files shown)


pkg/sharing/dockerconfigjson_test.go
```diff
@@ -63,4 +64,53 @@ func Test_NewCombinedDockerConfigJSON(t *testing.T) {
...
+			&corev1.Secret{
+				Data: map[string][]byte{
+					corev1.DockerConfigJsonKey: []byte(`{"auths":{"valid-server2":{"username":"user2","password":"pass2","auth":"auth2"},"empty-server2":{"username":"","password":"","auth":""}}}`),
+				},
+			},
+		}
+		result, err := NewCombinedDockerConfigJSON(secrets)
+		require.NoError(t, err)
+
+		// Empty entries should be skipped, only valid entries should be included in merged result
```
joaopapereira: This feels like a very verbose check which I think it makes it more complicated that it should be

joaopapereira: I would prefer to have just a Equal compare or something similar because this is just to much. If you want something that can output more verbose you could just convert the output to JSON and after that do a JSONEq

joaopapereira (changes requested):
Thanks for the PR. Can you please change the test as suggested?

pkg/sharing/dockerconfigjson_test.go
devacts: Coalesced all checks into 1 - to check for equality assertion with expected JSON

devacts:
@joaopapereira suggested change has been implemented. PTAL
Do let me know if the tests are enough or it requires more checks

praveenrewar (approved):

PR #750, opened 2026-09-22
Normalize registry hostname case when merging dockerconfigjson auths

sameerforge:
## Summary

- Normalize registry hostname casing when merging `dockerconfigjson` auths, so the documented "last secret wins" merge behavior holds regardless of case.
- Fix CI: stop bootstrapping a Buildx builder for the default kbld dev build, which was failing under minikube's docker-env.

## Problem

`NewCombinedDockerConfigJSON` merges `auths` entries from multiple `kubernetes.io/dockerconfigjson` Secrets, keyed by the raw registry hostname string. Registry hostnames are case-insensitive (RFC 1035/1123), but the merge treats `Harbor.Example.COM` and `harbor.example.com` as two distinct keys. If two Secrets reference the same registry with different casing, both entries survive in the combined map instead of the later, more-specific Secret overriding the earlier one — breaking the function's documented "last one wins" contract and potentially leaving a stale/duplicate credential in the combined dockerconfigjson.

## Fix

Normalize the registry key with `strings.ToLower()` before inserting it into the combined map (`pkg/sharing/dockerconfigjson.go`), so a later Secret's entry for the same registry always overrides an earlier one regardless of casing. Added a regression test (`pkg/sharing/dockerconfigjson_test.go`) covering two Secrets for the same registry in different cases, asserting only the later one survives.

## CI note

This PR's CI run was independently failing before it ever reached the unit tests: `hack/deploy.sh`'s kbld build bootstraps a Docker Buildx builder, and that builder's boot/inspect step 404s under minikube's containerd-backed `docker-env`. This wasn't about a moving/broken image tag — pinning the builder to an immutable `moby/buildkit` version hit the identical 404, which showed the actual root cause is Buildx's builder-container bootstrap itself being incompatible with this environment, not the image reference.

Looked closer and found `config/dev/build.yml`'s default values (`rapid_deploy=false`, `push_images=false` — exactly what CI uses) render an empty `docker.buildx: {}` stanza, which still routes the build through Buildx for zero actual benefit (no extra flags are ever applied in that case). Restructured the config so the default and `rapid_deploy` paths use kbld's classic `docker.build` (no Buildx involved), and only the `push_images` path — which genuinely needs multi-arch support — keeps `docker.buildx`. This also fixes a latent bug where `rapid_deploy`'s `buildkit: true` field was silently ignored (it's a valid field on the classic build options, not on the Buildx options, so it was being dropped by the type that was in use).

## Testing

- `go build ./...`
- `go test ./pkg/...` — all passing, including the new regression test
- `go vet ./...` — only pre-existing, unrelated warnings
- Verified the `config/dev/build.yml` change with `ytt` against all three value combinations (default, `rapid_deploy=true`, `push_images=true` with `platform` set) to confirm the `push_images`/multi-arch path is unchanged and only the default/CI path stops using Buildx


himsngh (approved):
LGTM!

